PR #1, opened 2015-04-07
Smack futur

jobol:
Hi Rob, no pressure, this pull request is not to be merged ;)

But please can you take a look to to all this commits that are implementing the security options for handling the LSM SMACK used at least by Tizen.

I will really appreciate your advises before re-working on it to be able to merge it. Maintaining separate commits is a plain and I would like to see all this work integrated upstream.


jobol:
too old and based on bad code. Should be reworked.

PR #2, opened 2015-04-08
mkdir: effective options -v and -p

jobol:
Despite the previous comment stating that options
of mkpathat are line up with -p and -v (or conversly),
options -p and -v wasn't effective for command 'mkdir'.

This patch make it effective.

Change-Id: I8d908eb852190b83d245e51e6d3f3ebdbfba7665
Signed-off-by: José Bollo <jose.bollo@open.eurogiciel.org>


jobol:
The problem solved here is an artefact of adding options to the right.

PR #6, opened 2015-06-10
sort: Make -z give null terminated output (to be like GNU coreutils)

Somasis:
I think I brought this up in IRC some time ago but nothing ever happened.


Somasis:
Sending to the mailing list instead.

PR #8, opened 2015-06-12
Fix ext3/4 blkid test comparisons

jbergstroem:
These two tests failed for me. Looks like a simple typo (and passes with the change).


jbergstroem:
Do I resend this to the ML or would you consider merging commit from here?

landley:
Resending to the mailing list would be easiest. (I'm traveling right now
and have access to the web archive, but not my email archive.)

Thanks,

Rob

jbergstroem:
Fixed in https://github.com/landley/toybox/commit/2a3ac1459f857f17e1c273c49a8df000b14fc8db.

PR #9, opened 2015-06-12
Fix a few clang warnings

jbergstroem:
Fixed a few warnings that's been annoying me. The majority of warnings left are related to types used in `printf()` (`-Wformat`).


Somasis:
Just a heads up that you should probably send these to the mailing list, I asked on IRC and that's the preferred place for patches

PR #14, opened 2015-09-08
increase the size of progname to avoid buffer overflow in netstat

LipiLee:
[ERROR]
root@generic_x86_64:/ # netstat -p
netstat: '1425/com.android.systemui' > 21 bytes


LipiLee:
Close this pull request and use mailing list.

PR #15, opened 2015-10-14
make.sh: missing cflags for config2help

Ajoyacr:
config2help might require extra parameters to be passed to the compiler. 
For example it might need an additional header path.


pohly:
I think this PR has been superseded by d20b9ff3ee0f. I have no problems compiling 0.7.0 unmodified in the same cross-compile environment as @Ajoyacr when settting HOSTCC.

PR #16, opened 2015-11-17
toybox: blkid: Properly distinguish ext2/ext3/ext4

tdm:
Current code relies on ext2 being at index 0 in the fstype array.  That
assumption was broken with upstream commit 1203ddf.  Instead, let's
compare against the fstype name so we don't rely on ordering.

Change-Id: Ie59ae9e85e5070f24e2069e57b660eac9e08a3a3


landley:
Sorry for the delay. Good catch, thanks. I fixed it another way (put it
back at the start where the comment said it should go) and changed the
test suite, which was expecting the broken behavior.

PR #19, opened 2016-02-08
stuff you may want...

izabera:


landley:
Fixed in a different way by commit 9448c3394465.

PR #20, opened 2016-02-09
main.c: fix non-root usage when installed suid root

pohly:
When toybox is installed suid root and invoked by a non-root user for
commands which do not require root privileges, it drops the root
privileges during initialization.

However, since commit afba5b8efd the result check of setuid() was
inverted such that it aborted on success, making toybox unusuable for
non-root users.

Signed-off-by: Patrick Ohly patrick.ohly@intel.com


landley:
applied

PR #21, opened 2016-02-09
toybox: tar: Fix support for long names

tdm:
Move the code to set hdr.name down to where the longname check is
performed so that toybox doesn't bail in xstrncpy().

Change-Id: Id025891993746889564b479e5185cf9721b54a55


tdm:
Will update the commit message and add a test shortly.

PR #23, opened 2016-02-11
removed unread assignment in sed.c

lovelycuppatea:
clang scan-build flags up 'to' as being unread, so clearing from code.


landley:
Applied.

PR #26, opened 2016-03-03
Fix ip build failure against musl

dimkr:
Signed-off-by: Dima Krasner dima@dimakrasner.com


enh-google:
(i'm assuming this is fixed, since rob builds against musl.)

PR #28, opened 2016-03-31
Rename basename_r to basename_readonly to avoid Bionic and FreeBSD conflict

emaste:
As suggested in the "FreeBSD porting, removing bashisms" thread.


emaste:
superseded by 1c028ca33dc059a9d8f18daafcd77b5950268f41

PR #32, opened 2016-04-20
Fix "touch -t" seconds parsing

DavisNT:
Seconds and nanoseconds parsing for `touch -t` was not working.

Command like `touch -t 200902132331.30 walrus` led to errors like `touch: bad '200902132331.30'`. 

This was caused by two factors:
* When seconds were present, but nanoseconds were **not** present `sscanf(s += len, "%lu%n", &ts->tv_nsec, &len)` was called, but did not set `len` to proper zero. This led to `s` being increased beyond the null terminator of the string by the further code.
* When seconds and nanoseconds were present `s` was not decreased (although `len` was increased for proper trailing zeros calculation by further code). This led to `s` being increased beyond the null terminator of the string by the further code.

In addition `touch -t` without seconds did **not** set seconds to zero. This was fixed as well.

Additionally tests for `touch -t` were added.

### DIFF
--- a/tests/touch.test
+++ b/tests/touch.test
@@ -11,6 +11,18 @@ testing "-c" "touch -c walrus && [ -e walrus ] && echo yes" "yes\n" "" ""
 testing "-c missing" "touch -c warrus && [ ! -e warrus ] && echo yes" \
   "yes\n" "" ""
 
+testing "-t" \
+  "touch -t 201201231234 walrus && date -r walrus +%Y%m%d-%H%M%S.%N" \
+  "20120123-123400.000000000\n" "" ""
+
+testing "-t seconds" \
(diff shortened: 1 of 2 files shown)


DavisNT:
@landley Could you please do the review? If any further improvements are needed in this code, please let me know! Thank you!

landley:
On 04/30/2016 12:56 PM, Dāvis Mošenkovs wrote:
> @landley <https://github.com/landley> Could you please do the review? If
> any further improvements are needed in this code, please let me know!
> Thank you!

I already applied this last week, it's commit 00385e8d0e1f.

(I wget the 32.patch file and then "git am" it. No merge commit that way.)

Rob

DavisNT:
Thank you! 
I see it in the code. :smile: Then I will close this pull request (as it has been merged).

Br,
Davis

PR #33, opened 2016-06-09
Fix dangling link after cp command

andyincode:
"toybuf" should be initialized before calling symlinkat() or
dangling link could occur.


landley:
Applied.

landley:
Applied long ago, and then fixed more generally in commit 46409d50e563.

PR #38, opened 2016-07-23
Fix memory leak in id found by LSan

jolfzverb:
Full LSan log:
```
$ ./generated/unstripped/toybox id -u
1000

=================================================================
==32119==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 80 byte(s) in 1 object(s) allocated from:
    #0 0x7f331a5c3ab2 in malloc (/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/libasan.so.2+0x98ab2)
    #1 0x411b20 in xmalloc lib/xwrap.c:64

Indirect leak of 400 byte(s) in 5 object(s) allocated from:
    #0 0x7f331a5c3ab2 in malloc (/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/libasan.so.2+0x98ab2)
    #1 0x411b20 in xmalloc lib/xwrap.c:64

SUMMARY: AddressSanitizer: 480 byte(s) leaked in 6 allocation(s).
```

### DIFF
--- a/toys/posix/id.c
+++ b/toys/posix/id.c
@@ -73,7 +73,8 @@ static void s_or_u(char *s, unsigned u, int done)
   else printf("%u", u);
   if (done) {
     xputc('\n');
-    exit(0);
+    toys.exitval = 0;
+    xexit();
   }
 }
 
@@ -143,7 +144,8 @@ static void do_id(char *username)
     }
     if (toys.optflags&FLAG_G) {
       xputc('\n');
-      exit(0);
+      toys.exitval = 0;
+      xexit();
     }
   }
 


landley:
Short-lived commands don't necessarily free their memory because exit()
will do it for them. We only worry about allocations building up,
perhaps based on size of input (one leak per line of sed or grep would
be bad if they're working on millions of lines of input), or based on
duration of running (leaking every 30 seconds in a server that runs days
would be bad). (Note that those allocations don't have to be _leaks_ to
be concerning.)

See the config symbol TOYBOX_FREE for details. (Config.in line 98 if you
don't want to deal with menuconfig).

Thanks,

Rob

jolfzverb:
Thank you!
Closing pull request now.

PR #42, opened 2016-08-11
Fix test applet issues

d4rken:
If the "!" operator was not used in the `test` applets argument,
`int not` was never initialized before using it in `toys.exitval ^= not`,
which lead to inconsistent results (e.g. unexpected result inversion).

### DIFF
--- a/toys/pending/test.c
+++ b/toys/pending/test.c
@@ -41,7 +41,7 @@ config TEST
 
 void test_main(void)
 {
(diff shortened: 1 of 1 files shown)


d4rken:
Second commit fixes the `!` operator being ignored if we return early due to the target not existing.

d4rken:
Sending this to the mailing list.

Issue #45, opened 2016-08-14
chown/chgrp aborts on numeric ids if they can't be resolved


d4rken:
From https://github.com/landley/toybox/issues/36#issuecomment-234476275

> I just noticed that "chown 123:123 toybox" is saying:
> chown: user '123'
> Which is broken in a couple ways (there should be a ":errno message" on the end of that if it's an actual error, and numeric user IDs should work even if they're not in /etc/passwd) so I need to off and fix THAT before cycling back to this. :)

chown uses [`xgetpwnamid`](https://github.com/landley/toybox/blob/8b685bae21f8753925dbefb8b07b78c47534f943/lib/xwrap.c#L593) (chgrp similar) to resolve the command input independent of whether it is a `uid` or a name, but if the function fails to resolve it throws an error and we exit.

Idea:
If `getpwnam` returns something we have a uid we can use.
if `getpwnam` returns null the input  can't be resolved to a uid. So it's either a uid or a user we can resolve. In this case we try turning the input into a number and use that as uid.
Same solution for group.

I'll try something and send a patch to the mailing list.

d4rken:
Fixed in https://github.com/landley/toybox/commit/3d64b0cc95c5957e53474c3e50f143c61a057104

PR #46, opened 2016-08-14
Allow processing of unknown uid/gid.

d4rken:
`xgetpwnamid` and `xgetgrnamid` only error out if the input is not a number.

(also submitted to mailing list)


d4rken:
#45

d4rken:
Fixed in https://github.com/landley/toybox/commit/3d64b0cc95c5957e53474c3e50f143c61a057104

PR #53, opened 2016-10-31
Fix control flow integrity check failure in ps.

eugenis:
This fixes an indirect function call through a pointer of an
incompatible type. An alternative is to change show_process pointer type
to void (*)(struct carveup *), but that requires a forward declaration of struct carveup
at the _global scope_, which is hard to do with the way globals.h is generated.

See http://clang.llvm.org/docs/ControlFlowIntegrity.html for more
details.

### DIFF
--- a/toys/posix/ps.c
+++ b/toys/posix/ps.c
@@ -544,8 +544,9 @@ static char *string_field(struct carveup *tb, struct strawberry *field)
 }
 
 // Display process data that get_ps() read from /proc, formatting with TT.fields
-static void show_ps(struct carveup *tb)
+static void show_ps(void *p)
 {
+  struct carveup *tb = p;
   struct strawberry *field;
   int pad, len, width = TT.width, abslen, sign, olen, extra = 0;
 
@@ -1215,7 +1216,7 @@ void ps_main(void)
   // print headers now (for low memory/nommu systems).
   TT.bits = get_headers(TT.fields, toybuf, sizeof(toybuf));
   if (!(toys.optflags&FLAG_M)) printf("%.*s\n", TT.width, toybuf);
-  if (!(toys.optflags&(FLAG_k|FLAG_M))) TT.show_process = (void *)show_ps;
+  if (!(toys.optflags&(FLAG_k|FLAG_M))) TT.show_process = show_ps;
   TT.match_process = ps_match_process;
   dt = dirtree_read("/proc",
     ((toys.optflags&FLAG_T) || (TT.bits&(_PS_TID|_PS_TCNT)))
@@ -1671,8 +1672,9 @@ static void do_pgk(struct carveup *tb)
   }
 }
 
-static void match_pgrep(struct carveup *tb)
+static void match_pgrep(void *p)
 {
+  struct carveup *tb = p;
   regmatch_t match;
   struct regex_list *reg;
   char *name = tb->str+tb->offset[4]*!!(toys.optflags&FLAG_f);;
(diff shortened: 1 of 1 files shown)


landley:
On 10/31/2016 02:34 PM, Evgeniy Stepanov wrote:
> This fixes an indirect function call through a pointer of an
> incompatible type. An alternative is to change show_process pointer type
> to void (*)(struct carveup *), but that requires a forward declaration
> of struct carveup
> at the /global scope/, which is hard to do with the way globals.h is
> generated.

How is this a bug again? The new code has to be able to typecast void *
to struct carveup * to use them. Are you saying that void * and struct
carveup pointer will _ever_ be different representations on the stack on
any Linux system?

You seem to be implying that C++ disease has advanced so far in modern
compilers that EVERY pointer ever passed in any function has to be void
* and then copied into a local variable of the appropriate type (with
implicit typecast on that assignment) to avoid triggering ever-expanding
"undefined" behavior.

I'm aware that when C++ developers took over C compiler development and
started reimplementing them in C++, they decided to sabotage C to bring
it down to parity with C++. C being a portable assembly language gave it
a significant advantage over C++, which is a minefield of magic behavior
you just have to memorize without rhyme or reason, constantly changing
every 3 years or so when they come out with a new 'standard', and if C++
developers allowed the same C program to compile for decades unmodified
when C++ can't do it for 3 years, then their argument that "C++ contains
the whole of C and is thus just as good a language, the same way a mud
pie contains an entire glass of water and is thus as good a beverage"
was undermined.

I'm aware they did that. And it is a BAD THING.

> See http://clang.llvm.org/docs/ControlFlowIntegrity.html for more
> details.

Uh-huh.

> Clang includes an implementation of a number of control flow
> integrity (CFI) schemes, which are designed to abort the program upon
> detecting certain forms of undefined behavior that can potentially
> allow attackers to subvert the program’s control flow.

How. How does this allow an attacker to subvert the control flow of a
binary calling its own internal functions? What's the attack?

I'm curious now. Please convince me this is a real issue.

Rob

eugenis:
Basically, these days code pages are not writable and stack/data pages - not executable, and the majority of exploits require finding a code pointer in writable memory and replacing it with something different. I.e. hijacking the control flow of the program. Of course this requires the ability to write to the program memory, but that's what all the heap overflow and use-after-free bugs are for.

CFI is about limiting possible control flow as much as possible. For function pointers it means run-time checking that the dynamic type of the call target matches the static type of the call site, i.e. calls though void (*)(void) pointer can only go to address-taken functions of type void(*)(void). This requires a bit of type discipline, but I'm pretty sure by ISO C standard calls through mismatching pointer type are undefined.

As I mentioned, this could be fixed in a number of different way, the best, probably, by changing the type of show_process to void (*)(struct carveup*). I could not do that w/o adding new warnings to the build though.

landley:
On 11/01/2016 12:56 AM, Evgeniy Stepanov wrote:
> Basically, these days code pages are not writable and stack/data pages -
> not executable, and the majority of exploits require finding a code
> pointer in writable memory and replacing it with something different.
> I.e. hijacking the control flow of the program. Of course this requires
> the ability to write to the program memory, but that's what all the heap
> overflow and use-after-free bugs are for.
> 
> CFI is about limiting possible control flow as much as possible.

So it's yet another layer of runtime checking trying to move the
exploits around.

Sigh, I'll throw it on the pile. Thanks for the heads up,

Rob

landley:
Applied.

PR #62, opened 2017-02-26
Fixes resource leak in netstat

usta:
Forgotten to close file before return statement

### DIFF
--- a/toys/net/netstat.c
+++ b/toys/net/netstat.c
@@ -115,7 +115,10 @@ static void show_ip(char *fname)
      return;
   }
 
-  if(!fgets(toybuf, sizeof(toybuf), fp)) return; //skip header.
+  if(!fgets(toybuf, sizeof(toybuf), fp)) { //skip header.
+     fclose(fp);
+     return;
+  }
 
   while (fgets(toybuf, sizeof(toybuf), fp)) {
     char lip[256], rip[256];


3bsilversmiths:
Ok

usta:
ping ?!

usta:
@landley

landley:
I hadn't bothered because A) it's not called in a loop (show_ip is called at most 6 times before the command exits, so filehandle exhaustion is highly unlikely to happen), B) it's a "should never happen, your kernel is broken" error return path that doesn't really need to be checked (the loop skips unparseable entries enyway). The same issue exists in show_unix_sockets() and display_routes() but you only addressed one. Ideally I'd like to convert both to use do_lines().

This command could actually use rather a lot of cleanup, but it's not one I use normally and hasn't got an automated regression test (nor is it easy to add one since it tests machine-variable network configuration), so adding such a test is part of the mkroot todo heap where I set up a synthetic system under qemu with known attributes for it to test and get consistent answers for. And that's on the post-1.0 todo list...

enh-google:
since it's Saturday and Larry can't tell me what to do, i've sent a patch to the mailing list. i didn't use do_lines() because that seemed more complicated than just using fscanf() directly rather than fgets() into toybuf and then sscanf(). i also fixed a bunch of formatting issues found when comparing the output to net-tools netstat, and a correctness issue with `netstat -xl`.

PR #82, opened 2018-01-21
Fix variables in help.txt

reverendhomer:
After replacing `echo TXT` with `cat FILE` in Makefile we don't need double dollar signs anymore.


reverendhomer:
Thanks for merging :)

PR #85, opened 2018-03-11
Add bc to pending

gavinhoward:
This bc is passing all math tests, and it successfully runs the [Linux timeconst.bc](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/time/timeconst.bc) script for all values from 0 to 10000, which is all that that particular script should need.

It is not done yet. I am planning to fuzz this for two weeks starting next week, and then I need to test it on more scripts as well. That's why I would like this in pending, not in posix. I am hoping to get it in posix by the next toybox release, though.

### DIFF
--- a/lib/args.c
+++ b/lib/args.c
@@ -76,8 +76,7 @@
 //     <0 die if less than # leftover arguments (default 0)
 //     >9 die if > # leftover arguments (default MAX_INT)
 //     ? Allow unknown arguments (pass them through to command).
-//     & first argument has imaginary dash (ala tar/ps)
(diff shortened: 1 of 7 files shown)


gavinhoward:
Oh, I forgot to say why I put everything in one file.

The only thing that I could think of that other stuff in toybox might use is the vector (resizable array). However, it is **really** small code, so I didn't see any benefit to splitting it.

It is just under 10,000 lines, though.

gavinhoward:
I accidentally pulled master, not merged. Will reopen.

PR #86, opened 2018-03-12
Add bc to pending (second attempt)

gavinhoward:
This is my second attempt to add bc to pending. I accidentally pulled from toybox and did not merge, which polluted the history.

This bc has been tested and found to work on basic math, as well as the Linux [timeconst.bc](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/time/timeconst.bc) script. However, it has not been thoroughly tested. Once it has, I will ask to move it to posix.

I decided to put all of the code into one file because there was only a small bit of code (vectors, aka resizable arrays) that might be useful to other commands. Yes, it is (just under) 10,000 loc. However, it must be remembered that this implements a complete programming language, including an interpreter.

### DIFF
--- a/toys/pending/bc.c
+++ b/toys/pending/bc.c
@@ -1,0 +1,9323 @@
+/* bc.c - An implementation of POSIX bc.
+ *
+ * Copyright 2018 Gavin D. Howard <yzena.tech@gmail.com>
+ *
+ * See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/bc.html
+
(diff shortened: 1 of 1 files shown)


gavinhoward:
In answer to an email to the toybox mailing list (that I didn't get because I was not subscribed to the mailing list at the time) to the previous (now defunct) PR, , this bc passes all math tests for +, -, *, /, %, ^, and sqrt.

I don't know how to set up tests for bc in toybox. I did, however, write my tests so that I could test the bc when integrated into toybox. I could add tests if you would like.

gavinhoward:
I will also close this PR because I hadn't realized that you had already pulled the code in. (GitHub did not tell me for some reason.)

PR #89, opened 2018-04-19
Accept OACK when requesting a file.

filipperich:
Optionally, accept and acknowledge OACK from server after sending original RRQ, if the server implements RFC2347 and responds first with an OACK before sending DATA packets.


landley:
Applied.

PR #92, opened 2018-06-02
call strip as $STRIP

WallaceIT:
This is useful e.g. for cross toolchains that define $CC and $STRIP
instead of $CROSS_COMPILE.


landley:
applied

PR #94, opened 2018-06-10
toysh: add exec nonfork command

WallaceIT:


WallaceIT:
This is required also for correct execution of _switch_root_  in case _toysh_ (or a bash script using it) is used as init process.

landley:
exec was added in commit b876221b7f07

PR #95, opened 2018-06-21
wget: use error_exit if file already exists

reverendhomer:
With `perror_exit`, `wget` fails with `"file already exists: No such file or directory"`

### DIFF
--- a/toys/pending/wget.c
+++ b/toys/pending/wget.c
@@ -135,7 +135,7 @@ void wget_main(void)
 
   // TODO extract filename to be saved from URL
   if (!(toys.optflags & FLAG_f)) help_exit("no filename");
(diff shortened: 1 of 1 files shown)


landley:
Fixed a slightly different way in commit 5e940c3806f5

reverendhomer:
Hmm, are you sure you actually *fixed* this? After your commit, the message will look like
```
'foo' already exists: No such file or directory
```
because of `perror_exit`

landley:
On 08/03/2018 10:39 AM, Reverend Homer wrote:
> Hmm, are you sure you actually /fixed/ this? After your commit, the message will
> look like
> 
> |'foo' already exists: No such file or directory |
> 
> because of |perror_exit|

Sigh. I converted perror->error locally, and then reverted it accidentally. My bad.

Rob

PR #47, opened 2016-08-14
Fixes for test applet NOT operator

d4rken:
(also submitted to mailing list)

PR #100, opened 2018-08-28
test: Fix negation for comparison operators

baileyforrest:
The value is negated on the last line to reflect unix return code:
toys.exitval ^= not

### DIFF
--- a/toys/pending/test.c
+++ b/toys/pending/test.c
@@ -94,10 +94,10 @@ void test_main(void)
       s = toys.optargs[1] + 1;
       if (!strcmp("eq", s)) toys.exitval = a != b;
       else if (!strcmp("ne", s)) toys.exitval = a == b;
-      else if (!strcmp("gt", s)) toys.exitval = a < b;
-      else if (!strcmp("ge", s)) toys.exitval = a <= b;
-      else if (!strcmp("lt", s)) toys.exitval = a > b;
-      else if (!strcmp("le", s)) toys.exitval = a >= b;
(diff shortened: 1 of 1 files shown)


d4rken:
Is this related to #47?

baileyforrest:
I don't think so. #47 assign toys.exitval:
https://github.com/landley/toybox/pull/47/commits/e3530589637c19c6e2f8c761a3efaf0e3ab0f936

The comparison operators all overwrite this value:
https://github.com/landley/toybox/pull/100/commits/6d6a08bfaf20f87008bcc3fc182c861952407219#diff-b1283e40e24032202900ace3816856d8L95

The main point of this PR is the value returned for comparison operators is incorrectly negated.

This is supposed to print "good", however with toybox it prints "bad":
if test "foo" = "foo"; then echo good; else echo bad; fi

spiterikevin:
It might be related because toys.exitval should be 0 for true and 1 for false. Testing #100 without applying #47 first might give opposite results because the uninitialized not can be messing up the result in line 112.

Lines 95-100 should have inverse logic. I think 95,96 should be unchanged, and 97-100 should be corrected to <=, <, >=, > respectively.

baileyforrest:
@spiterikevin , you are correct. I did not realize that 'not' was uninitialized. Apparently on our platform it is always '1'.

@d4rken , this does depend on #47 now.

enh-google:
test.c was rewritten. if you have a specific test case (no pun intended) that still fails, let us know...

PR #102, opened 2018-08-30
test.c crashes with incorrect number of arguments

spiterikevin:
There was a segment violation checking a non-existing parameter for "!". The argument count was also not handled correctly in other places.


spiterikevin:
Resolved by ef3044114acac33c8231ace80766ec6d9133150a.

Issue #108, opened 2018-12-11
cp segfault with empty list in --preserve


BookGin:
## Description
When specify `--preserve` without providing a list, `cp` will segfault.

```sh
$ cp --preserve foo bar
[1]    16649 segmentation fault (core dumped)  /tmp/toybox/bin/cp --preserve foo bar
```

The root cause is `xstrdup(TT.c.preserve)`, because `TT.c.preserve` is NULL here. 

https://github.com/landley/toybox/blob/master/toys/posix/cp.c#L378

## Version

```sh
$ cp --version
toybox android-wear-8.0.0_r2-1-gb5744abfe599
```

landley:
Fixed in commit d4017dbd0a42. Thanks for the heads up.

PR #109, opened 2018-12-11
cp: fix --preserve segfault bug

BookGin:
#108


landley:
I implemented the "default = mot" value from the ubuntu man page instead. (That's why the optstr had the semicolon, to allow that.)

Issue #103, opened 2018-09-03
memory issue in config2help (heap-use-after-free)


leeyh8901:
There is memory error (heap-use-after-free) issue in config2help. 
For reproduce issue,  I had compiled below command. (script/make.sh)

# Issue
(1) compile config2help with asan
$ gcc -g -fsanitize=address scripts/config2help.c -o config2help

(2) Run config2help
$ config2help Config.in .config
<pre><code>
#define HELP_mkdir "usage: mkdir [-vp] [-m mode] [dirname...]\n\nCreate one or more directories.\n\n\n"

=================================================================
==8566== ERROR: AddressSanitizer: heap-use-after-free on address 0x60060000eb10 at pc 0x4046a1 bp 0x7ffe439ac820 sp 0x7ffe439ac818
READ of size 8 at 0x60060000eb10 thread T0
    #0 0x4046a0 (/home/yhji/Works/toybox/toybox/config2help+0x4046a0)
    #1 0x7fb891d00ec4 (/lib/x86_64-linux-gnu/libc-2.19.so+0x21ec4)
    #2 0x401078 (/home/yhji/Works/toybox/toybox/config2help+0x401078)
0x60060000eb10 is located 16 bytes inside of 24-byte region [0x60060000eb00,0x60060000eb18)
freed by thread T0 here:
    #0 0x7fb8920ba33a (/usr/lib/x86_64-linux-gnu/libasan.so.0.0.0+0x1533a)
    #1 0x401a35 (/home/yhji/Works/toybox/toybox/config2help+0x401a35)
    #2 0x403c86 (/home/yhji/Works/toybox/toybox/config2help+0x403c86)
    #3 0x7fb891d00ec4 (/lib/x86_64-linux-gnu/libc-2.19.so+0x21ec4)
previously allocated by thread T0 here:
    #0 0x7fb8920ba41a (/usr/lib/x86_64-linux-gnu/libasan.so.0.0.0+0x1541a)
    #1 0x401154 (/home/yhji/Works/toybox/toybox/config2help+0x401154)
    #2 0x4016d5 (/home/yhji/Works/toybox/toybox/config2help+0x4016d5)
    #3 0x402659 (/home/yhji/Works/toybox/toybox/config2help+0x402659)
    #4 0x402d68 (/home/yhji/Works/toybox/toybox/config2help+0x402d68)
    #5 0x7fb891d00ec4 (/lib/x86_64-linux-gnu/libc-2.19.so+0x21ec4)
[...]

PR #110, opened 2018-12-12
Fix memory issue in config2help

leeyh8901:
- issue#103
- https://github.com/landley/toybox/issues/103
- heap-use-after-free
- Change to find double dash (CP_PRESERVE)


landley:
Commit 185722d30ae7 removed config2help (replaced by -h option of new 
kconfig.c).

PR #113, opened 2019-01-08
Intial FreeBSD support

emaste:
These four changes provide initial support for FreeBSD. The first three are final, while the workaround is a bit of a hack.

With this change I can build these posix toys:
```
% ./toybox
basename cal cat catv chgrp chmod chown cksum cmp comm cpio crc32
cut date dirname du egrep expand false fgrep file find grep groups
head iconv id kill killall5 link logger logname ls mkdir mkfifo nice
nl nohup od paste patch printf pwd renice rm rmdir sed sleep sort
split strings tee test time touch true tty uname uniq unlink uudecode
uuencode wc who whoami xargs
```

Notable missing POSIX toys: `cp mv install df getconf ln ps top iotop pgrep pkill tail`


landley:
If you can't trust bash to be at a known absolute path, how can you trust env to be at a known absolute path? I've never understood the point of this gratuitous layer of indirection...

landley:
I think I've now applied everything but the env stuff.

emaste:
> how can you trust env to be at a known absolute path?

Only that in practice every relevant system has a `/usr/bin/env`, and not every system has `/bin/bash`.

landley:
> Only that in practice every relevant system has a /usr/bin/env,

Android hasn't got a /usr, we'll need to make a posix build container. (It's been discussed on the list.)

> and not every system has /bin/bash.

MacOS did. What else is missing it?

emaste:
Hmm, interesting point about Android - I hadn't really considered Android as a build host before.

> What else is missing it?

*BSD will generally have /usr/local/bin/bash - FreeBSD, OpenBSD, NetBSD, DragonflyBSD.

landley:
Android as a build host is paragraph 2 of http://landley.net/toybox/about.html and then most of the video below that. (The 2013 "Why is toybox" talk it has a lot of section links into the middle of...)

emaste:
Hrm, this refreshed with my experimental WIP. I'll close this pull request and submit another with changes proposed for upstream.

PR #115, opened 2019-02-15
Fix typos: e.g. change "cannonical" to "canonical".

jasonspiro:
# Background

The actual runnable code has never misspelled "canonical" or "canonicalize", but some code comments (and www/news.html) have occasionally misspelled one or both words.

This pull request fixes the misspellings.

# Possible memory aid

Here's a **possible memory aid** which might help you in the future:

A "canonical" path is an official, authoritative path, as in "canon law".  A "cannonical" path is a trail you can use to transport heavy artillery which can shoot cannonballs at an enemy army.

So, canons improve your paths, but cannons explode and destroy them.

# Conclusion

Thank you for continuing to maintain `toybox`!  I still don't miss the old Android `toolbox` at all.

### DIFF
--- a/lib/xwrap.c
+++ b/lib/xwrap.c
@@ -504,7 +504,7 @@ void xstat(char *path, struct stat *st)
   if(stat(path, st)) perror_exit("Can't stat %s", path);
 }
 
-// Cannonicalize path, even to file with one or more missing components at end.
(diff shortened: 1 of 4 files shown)


landley:
Applied.

PR #116, opened 2019-02-15
route: fix command argument parsing

uprestor:
Running 'route add default gw X' fails with message `route: error resolving 'gw'`.  Argument parsing code in `get_next_params()` and `get_next_params_ipv6()` is using `*argv` for parameter value instead of `argv[1]`.

### DIFF
--- a/toys/pending/route.c
+++ b/toys/pending/route.c
@@ -201,36 +201,37 @@ static void get_next_params(char **argv, struct rtentry *rt, char **netmask)
 
       //set the metric field in the routing table.
       if (!strcmp(*argv, "metric"))
-        rt->rt_metric = atolx_range(*argv, 0, ULONG_MAX) + 1;
+        rt->rt_metric = atolx_range(argv[1], 0, ULONG_MAX) + 1;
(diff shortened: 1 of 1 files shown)


landley:
Thank you for this, but I have complete rewrite of this command pending based on the rtnetlink API. What's your use case, and would you be interested in testing the new version?

uprestor:
I am using toybox for an internal project, and I needed the route(1) functionality.  I opened the PR to make sure the fix gets propagated upstream.  Having said that if you are in the process of a rewrite I'd be happy to test the new version.  

Do you have a timeframe in mind for the new version?  If it is imminent I'd be happy to close this PR.  However, if you plan to release a new version before the rewrite would you mind merging this PR so that we don't loose the fix if we import a new version?

landley:
Applied, but I'm probably checking in a complete replacement... I'll see what I can have ready by next weekend, maybe?

(It's always awkward replacing an existing implementation. When is the new one good enough, and how can it be without testing? In _theory_ this is what pending is for, but people use things out of pending...)

Issue #124, opened 2019-03-25
find: no support for multiple -exec sections


treysis:
`find` should allow several uses of the `-exec` parameter. The current version ignores all following `-exec` parameters except for the first one.

Example:
`find . -type f -exec stat -c %n {} \; -exec stat -c %Y {} \;`
Should list all the files and their mtime in a directory. However, only the filename is reported. Works with busybox and on "regular" linux systems (e.g. Linux Mint).

landley:
Koalaman's fix appears to have fixed it for me, please confirm?

treysis:
Hi,

 

was there an easier way than cloning Koalaman's repo and building myself? Well, that's what I did at least (fortunately I just learned how to use git more or less over the weekend).

 

Result: works as intended now.

 

Thx for the fast response. Unfortunately, this probably won't make it into Android before Android 11...

 

Cheers,

T

 

Gesendet: Dienstag, 26. März 2019 um 16:01 Uhr
Von: "Rob Landley" <notifications@github.com>
An: landley/toybox <toybox@noreply.github.com>
Cc: treysis <treysis@gmx.net>, Author <author@noreply.github.com>
Betreff: Re: [landley/toybox] find: no support for multiple -exec sections (#124)


Koalaman's fix appears to have fixed it for me, please confirm?

enh:
merged into AOSP as https://android-review.googlesource.com/c/platform/external/toybox/+/933940, so should be in Q.

i think this can be closed now :-)

PR #125, opened 2019-03-25
find: fix inverted -exec exit status

koalaman:
The return value of -exec was the command's exit code, which did not
account for the fact that an exit code of zero means success, while in
C, zero means failure. From POSIX:

>  the primary shall evaluate as true if the utility returns a zero
>  value as exit status

This commit flips the return value, and adds two tests.

This fixes https://github.com/landley/toybox/issues/124


treysis:
Seems to work as intended now. Didn't encounter any misbehavior of the fix. 👍

PR #136, opened 2019-08-04
Make mktemp --tmpdir's argument optional

E5ten:
This matches GNU mktemp which has an optional argument for --tmpdir


landley:
Except it applies to both the short and longopt.

$ mktemp -p
mktemp: option requires an argument -- 'p'
Try 'mktemp --help' for more information.
$ ./mktemp -p
bash: ./mktemp: No such file or directory

landley:
Ahem: that second one should obviously be:
$ ./mktemp -p
/tmp/tmp.XXXXqDjUXU

PR #138, opened 2019-08-28
tar not applying correct file attrs anymore

danil-ebali:
Seems like commit 8994b82280f0b562ed70d9ae38f1a5103b2791f2
introduced this bug on regular files


landley:
Was this fixed by 1a7c18fb818c or is it a separate issue?

landley:
Assuming this was fixed (in the absence of a test case).

PR #143, opened 2019-09-29
Add ln -t

E5ten:


landley:
What's your use case here?

I ask because you're the first person to want this flag. It's not in posix (https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ln.html (or LSB https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/cmdbehav.html) and busybox is 20 years old and still doesn't have this flag, presumably because nobody's ever asked them for it?

I'm happy to add features people need, but filling them in just because they're not there is "infrastructure in search of a user". (If a user for such a thing ever _does_ show up, there's no guarantee the feature meets their needs, and it's gone untested in the tree for long enough it may have bit-rotted anyway as the context of the project (and kernel, and libc, and compiler) shifted under it.

Since it's just as easy to add the thing later when there's a real user to provide immediate feedback, I kinda want to know the use case so I can at least add a real-world-inspired test to the test suite.

E5ten:
It's used by my distro for this: https://git.archlinux.org/svntogit/packages.git/tree/trunk/update-ca-trust?h=packages/ca-certificates and it's used within the build system for iwd. I can't think of other specific examples off the top of my head but I'm pretty sure there have been 1 or 2 other scripts that make use of it that I either had to modify or not use because of the `-t` use. At the moment I rebuild the package above locally with a modification to the script to not use `-t` but I'd prefer to just be able to use it as provided.

landley:
Applied 14f5f890021b0

PR #144, opened 2019-09-29
Add rmdir --ignore-fail-on-non-empty

E5ten:


landley:
This is a build break because - is an operator so using it in a FLAG() macro expands to a subtraction of multiple nonexistent variable names. Header generation converts - to _ in FLAG_* names (on line 109 of https://github.com/landley/toybox/blob/master/scripts/mkflags.c#L109 ) so if you use the FLAG(name_with_underscores_instead_of_minuses) version it should work.

Out of curiosity, what's your use case for this? (What script needs the -rmdir-ignore-fail-on-very-long-option-name-with-no-short-equivalent?)

E5ten:
Huh, my local patch has _s because I encountered the build failure you mentioned, I don't know how I committed the previous one instead. In terms of my use-cases: https://github.com/dell/dkms/blob/master/dkms#L1801 https://git.archlinux.org/svntogit/packages.git/tree/trunk/xorg-mkfontscale.script?h=packages/xorg-mkfontscale
I do agree that the option name is ridiculously long.

PR #145, opened 2019-10-03
Add ln -r

E5ten:
I had to make getdirname return "." instead of the file when the path has no '/', like dirname does, for my code to work correctly. I checked the new flag and toybox ln in general against GNU's ln tests.
Use cases:
https://git.archlinux.org/svntogit/packages.git/tree/trunk/update-ca-trust?h=packages/ca-certificates#n38
https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/linux#n201

### DIFF
--- a/toys/posix/ln.c
+++ b/toys/posix/ln.c
@@ -31,14 +31,6 @@ GLOBALS(
   char *t;
 )
 
-static int common_path(char *path1, char *path2)
-{
-  int i = 0;
-
-  while (*path1++ == *path2++) i++;
-  return i;
-}
-
 void ln_main(void)
 {
   char *dest = TT.t ? TT.t : toys.optargs[--toys.optc], *new;
@@ -68,23 +60,24 @@ void ln_main(void)
     else new = dest;
 
     if (FLAG(r)) {
-      char *abs, *s;
+      if (FLAG(s)) {
+        int j = 0, k;
+        // absolute paths needed for comparison
+        char *s = getdirname(new), *ss, *try2 = xabspath(try, -1)
+          *abs = xmprintf("%s/%s", ss = xabspath(s, -1), getbasename(new));
 
-      if (!FLAG(s)) error_exit("-r only with -s");
+        free(s);
+        free(ss);
 
-      // common_path needs absolute paths for comparison
-      try = xabspath(try, -1);
-      s = getdirname(new);
-      abs = xmprintf("%s/%s", xabspath(s, -1), getbasename(new));
-      free(s);
+        while (try2[j] == abs[j] && strchr(try2 + j, '/')) j++;
+        for(k = j, rc = 0; abs[k]; k++) if(abs[k] == '/') rc++;
+        free(abs);
 
-      // avoid stripping common path if source and target are the same file
-      if (strcmp(try, abs)) {
-        rc = common_path(try, abs);
-        try += rc;
-        abs += rc;
-        for (;*abs; abs++) if (*abs == '/') try = xmprintf("../%s", try);
-      }
+        try = xzalloc(strlen(try2) - j + rc * 3 + 1);
+        while (rc--) strcat(try, "../");
+        strcat(try, try2 + j);
+        free(try2);
+      } else error_exit("-r only with -s");
     }
 
     // Force needs to unlink the existing target (if any). Do that by creating
@@ -117,6 +110,7 @@ void ln_main(void)
                        FLAG(s) ? "symbolic" : "hard", try, new);
     else if (FLAG(v)) fprintf(stderr, "'%s' -> '%s'\n", new, try);
 
+    if (FLAG(r)) free(try);
     if (new != dest) free(new);
   }
 }
--- a/toys/posix/ln.c
+++ b/toys/posix/ln.c
@@ -4,13 +4,13 @@
  *
  * See http://opengroup.org/onlinepubs/9699919799/utilities/ln.html
 
-USE_LN(NEWTOY(ln, "<1t:Tvnfs", TOYFLAG_BIN))
+USE_LN(NEWTOY(ln, "<1t:Tvnfrs", TOYFLAG_BIN))
 
 config LN
   bool "ln"
   default y
   help
-    usage: ln [-sfnv] [-t DIR] [FROM...] TO
+    usage: ln [-sfnrv] [-t DIR] [FROM...] TO
 
     Create a link between FROM and TO.
     One/two/many arguments work like "mv" or "cp".
@@ -18,6 +18,7 @@ config LN
     -s	Create a symbolic link
     -f	Force the creation of the link, even if TO already exists
     -n	Symlink at TO treated as file
+    -r	Create symbolic links relative to link location
     -t	Create links in DIR
     -T	TO always treated as file, max 2 arguments
     -v	Verbose
@@ -29,6 +30,14 @@ config LN
 GLOBALS(
   char *t;
 )
+
+static int common_path(char *path1, char *path2)
+{
+  int i = 0;
+
+  while (*path1++ == *path2++) i++;
+  return i;
+}
 
 void ln_main(void)
 {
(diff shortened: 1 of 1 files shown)


landley:
The previous function always returned a malloc() you could free(), and the comment you modified said it ALWAYS returns a malloc. The new one sometimes returns a constant "." that you can't free(). These are not the same object lifetimes, and you can't easily tell when it's done what (if you send in a "." it will return "." but a _malloced_ ".").

I didn't use https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/dirname.html because A) it modifies the string it points to and B) there's no clarity about the lifetime of the object it returns. Instead I made something that always returns an allocation you can free.

E5ten:
Sorry about that, I'm incredibly new to C and didn't understand that there was another aspect to the benefit from getdirname than the lack of modification of what it points to. I'll try to find out how to do this correctly.

E5ten:
I think this should make it always return a malloc now while still returning "." for paths with no /

E5ten:
I guess that means I should free() my use of getdirname() after the fact? would this be right?
```
diff --git a/toys/posix/ln.c b/toys/posix/ln.c
index aa1b087f..e495a756 100644
--- a/toys/posix/ln.c
+++ b/toys/posix/ln.c
@@ -68,13 +68,14 @@ void ln_main(void)
     else new = dest;

     if (FLAG(r)) {
-      char *abs;
+      char *abs, *s = getdirname(new);

       if (!FLAG(s)) error_exit("-r only with -s");

       // common_path needs absolute paths for comparison
       try = xabspath(try, -1);
-      abs = xmprintf("%s/%s", xabspath(getdirname(new), -1), getbasename(new));
+      abs = xmprintf("%s/%s", xabspath(s, -1), getbasename(new));
+      free(s);
       rc = common_path(try, abs);
       try += rc;
       abs += rc;
```

lib/lib.c
```diff
@@ -1004,12 +1004,14 @@ void mode_to_string(mode_t mode, char *buf)
...
-// This always returns a malloc() copy of everyting before last (run of ) '/'.
+// This always returns a malloc() copy of everyting before last (run of ) '/',
+// or "." if there is no '/'
 char *getdirname(char *name)
 {
   char *s = xstrdup(name), *ss = strrchr(s, '/');
 
-  while (ss && *ss && *ss == '/' && s != ss) *ss-- = 0;
+  if (ss) while (ss && *ss && *ss == '/' && s != ss) *ss-- = 0;
+  else s = xstrdup(".");
```
jmakip: malloc from first xstrdup is never freed when else happens?

E5ten: I think this should be fixed but I don't want to resolve this until someone who knows what they're doing a bit better verifies it.

E5ten:
ping?

landley:
Sorry, looking at it today. It's the lib/ change that's slowing it down, if I'm changing the behavior of that I want it to work right for everybody and replace dirname() out of libc. (I substituted that one into dirname.c and the test suite failed because "/" returns "/" and... working on it.)

E5ten:
@landley 
```
char *getdirname(char *name)
{
  int i = 0;
  char *s = xstrdup(name), *ss = strrchr(s, '/');

  if (*++ss) i += 1;
  ss -= 1;
  if (ss) {
      while (ss && *ss && s != ss && *ss == '/' || i < 2) {
        if (*ss != '/') if (!i++) *ss = '.';
         else if (i < 2) *ss = 0;
        ss--;
      }
  } else {
    free(s);
    s = xstrdup(".");
  }

  if (i < 2) {
    free(s);
    s = xstrdup(".");
  } else if (!*s) {
    if (!i) free(s);
    s = xstrdup("/");
  }

  return s;
}
```
This is probably very overengineered and poorly done and I don't know for sure if it deals with the always returning a malloc correctly, or if all parts of it are necessary, and even if they are they very likely can be done more shortly, but it does pass all tests when substituted into dirname.c.

jmakip:
I dont understand your first if structure.
What happens if your strrchr returns 0 into ss. 
Then you  preincrement nullpointer and next line -= 1;

I would write it following way. 

.patch attatched as .txt file
[0001-lib-getdirname.txt](https://github.com/landley/toybox/files/3725006/0001-lib-getdirname.txt)

```
char *getdirname(char *name)
{
  char *s = xstrdup(name), *ss = strrchr(s, '/');

  if (!ss) free(s), xstrdup(".");
  else {
    ss = s+strlen(s)-1;
    if (ss != s) do {
      *ss-- = 0;
    } while(*ss != '/');
  }

  return s;
}

**Comment by E5ten** [comment]
I don't really understand some of what I did there, I was exhausted when I wrote it and it's super messy and gross and definitely broken so far. I don't think yours works either yet though because for me when subbed into dirname.c all of the tests failed.

**Comment by landley** [comment]
I did:

char *getdirname(char *name)
{
  char *s, *ss, *keep;

  for (s = name, ss = keep = 0; ; s++) {
    if (!*s) return keep ? xstrndup(name, keep-name) : xstrdup(".");
    if (*s != '/') keep = ss;
    else if (s == name) keep = ss = s+1;
    else if (s[-1] != '/') ss = s;
  }
}

And make test_dirname passed, I'm just trying to audit and think through the callers of dirname() out of libc (which the man page says has two versions and I don't want two versions because behavior changing due to libc environment weirdness is bad) to make sure I understand what substituting it in everywhere would mean...

Sorry, changing libc infrastructure can lead me down ratholes...

**Comment by landley** [comment]
Github's comments don't understand indentation, I see. Bravo, github. Never try to do python or makefiles.

**Comment by E5ten** [comment]
```
char *getdirname(char *name) {
  char *end = name + strlen(name) - 1;

  while (*end == '/' && end > name) end--;
  while (*end != '/' && end >= name) end--;
  while (*end == '/' && end > name) end--;

  return ++end == name ? xstrdup(".") : xstrndup(name, end - name);
}
```
I asked @morganamilo for help with this, all credit goes to her for writing the function.

**Comment by enh-google** [comment]
(you need to use Markdown syntax and surround the code with \`\`\` for blockquoting (which is what the examples above used), or if *every* line is indented then it's assumed. but since C code tends to start in column 0, you'll want ```.)

**Comment by E5ten** [comment]
My github didn't update and I didn't even see that one got posted 5 minutes before I commented.

**Comment by jmakip** [comment]
Tests in test_dirname looks fun. I just tested with these. Robs implementation is better but he should add null pointer check if want to behave what posix says

> If path is a null pointer or points to an empty string, dirname() shall return a pointer to the string "." .

```
  char* path0 = "/usr/lib/";
  char* path1 = "/usr/lib";
  char* path2 = "/usr/";
  char* path3 = "usr";
  char* path4 = "/";
  char* path5 = ".";
  char* path6 = "..";
  char* path7 = "";
  char* path8 = 0;

  printf("path  getdirname  \n");
  printf("%s  %s\n", path0, getdirname(path0));
  printf("%s  %s\n", path1, getdirname(path1));
  printf("%s  %s\n", path2, getdirname(path2));
  printf("%s  %s\n", path3, getdirname(path3));
  printf("%s  %s\n", path4, getdirname(path4));
  printf("%s  %s\n", path5, getdirname(path5));
  printf("%s  %s\n", path6, getdirname(path6));
  printf("%s  %s\n", path7, getdirname(path7));
  printf("%s  %s\n", path8, getdirname(path8));
  //printf("%s  %s\n", path8, dirname(path8));
```

E5ten:
@landley should I update this PR with working getdirname or are you going to add that to toybox yourself at which point I could rebase this PR over that?

landley:
My reply got long so I posted it to the mailing list. http://lists.landley.net/pipermail/toybox-landley.net/2019-October/011127.html

landley:
I broke down and implemented it myself. Sorry. (I'm a bit overwhelmed with incoming bug reports at the moment and I fly to Japan tuesday morning, so I'm trying to close tabs.)

PR #148, opened 2019-10-10
expr: fix memory management

### LINKED ISSUE #147: expr breaks autoconf
example command taken from autoconf with set -x enabled:
`expr 'x--trace=AC_CONFIG_HEADERS:$n::${::}%' : 'x--[^=]*=\(.*\)' \| 'x--trace=AC_CONFIG_HEADERS:$n::${::}%' : 'x-.\(.*\)'`
With other expr implementations, this outputs `AC_CONFIG_HEADERS:$n::${::}%`, while with toybox this outputs something wrong and different each time like `@S{1U`
This doesn't occur when the line `if (TT.refree) free(TT.refree);` in expr.c is removed, not to suggest that simply removing the line is the solution, because that does cause a leak, but just that it's where the problem is.

E5ten:
Andy Chu's patch from the mailing list, modified slightly to fit the current source.
http://lists.landley.net/htdig.cgi/toybox-landley.net/2016-March/008113.html

This fixes #147 as well as fixing all remaining memory management issues in expr from what I can see (valgrind now gives no warnings or any indication of issues while previously it reported many invalid reads).


E5ten:
ping

toys/pending/expr.c
```diff
@@ -255,7 +263,16 @@ void expr_main(void)
   if (ret.s) printf("%s\n", ret.s);
   else printf("%lld\n", ret.i);
 
-  toys.exitval = is_false(&ret);
+  int status = is_false(&ret);
+
+  // Free all the strings we allocated.
+  struct arg_list *h, *head = TT.allocated;
```
enh-google: use llist_traverse()?

(plus for some reason rob doesn't believe in C99, so declarations would have gone at the top.)

toys/pending/expr.c
```diff
@@ -255,7 +263,16 @@ void expr_main(void)
...
+  while (head) {
+    h = head;
+    head = head->next;
+    free(h->arg); // free the string
+    free(h); // free the node for tracking the string
+  }
 
-  if (TT.refree) free(TT.refree);
+  toys.exitval = status;
```
enh-google: why not just set `toys.exitval` directly on L266?

E5ten:
uh I did not mean to do that

enh-google:
(you might want to squash the two commits into one for ease of review. you probably also want to just send the patch on the list, because rob pays more attention to that than to the pull requests...)

PR #149, opened 2019-10-15
WIP: add -t to cp, mv, and install

E5ten:
I expect there will be issues within this, that said I did not find any immediately that I didn't fix before committing and I figured getting a WIP PR up would be the best way to expose further problems. In terms of use cases:
https://github.com/archlinux/mkinitcpio/blob/master/Makefile this uses cp -t
https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/linux and this uses both cp -t and install -t. I have not seen mv -t used, but saw no reason to leave it out for mv alone since GNU mv has it.

### DIFF
--- a/toys/posix/cp.c
+++ b/toys/posix/cp.c
@@ -15,15 +15,15 @@
 // options shared between mv/cp must be in same order (right to left)
 // for FLAG macros to work out right in shared infrastructure.
 
-USE_CP(NEWTOY(cp, "<2"USE_CP_PRESERVE("(preserve):;")"D(parents)RHLPprdaslvnF(remove-destination)fi[-HLPd][-ni]", TOYFLAG_BIN))
-USE_MV(NEWTOY(mv, "<2vnF(remove-destination)fi[-ni]", TOYFLAG_BIN))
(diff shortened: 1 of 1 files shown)


landley:
I'm confused... what changed? (The diff of the 149.patch I downloaded earlier and the 149.patch I just downloaded is... large.)

landley:
Is it just me or does:

+USE_MV(NEWTOY(mv, "<1t:vnF(remove-destination)fit:[-ni]", TOYFLAG_BIN))

Have t: in it twice?

landley:
+  if (!toys.optc && !TT.t) error_exit("needs 2 args");

So if you say "cp" by itself it'll say "Needs 1 argument", and then if you go "cp file" it'll say "needs 2 args" with no explanation, and different phrasing and capitalization. The current lib/args.c output is:

$ toybox cp blah
cp: Need 2 arguments (see "cp --help")

(Which comes from help_exit() rather than error_exit().)

Rob

P.S. There's a reason I generally do this stuff on the mailing list. You're gonna get a lot of short emails from this...

enh-google:
3d5cb06e8b5cfcd4226375ff435b1725399fda13 adds -t to install, but not cp/mv.

landley:
Commit 6d615c73b321 then.

PR #150, opened 2019-10-15
Add -c to chmod, improve message, and switch to FLAG() macros

E5ten:
Use case: https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/linux


landley:
Commit df570102af78

PR #156, opened 2019-12-05
Fix getopt optstring

E5ten:
Move longopts after their corresponding shortopts instead of before

### DIFF
--- a/tests/getopt.test
+++ b/tests/getopt.test
@@ -30,7 +30,7 @@ test_getopt "-o a -l one:: -- -a --one arg" " -a --one '' -- 'arg'"
 test_getopt "-o a -l one:: -- -a --one=arg" " -a --one 'arg' --"
 
 # "Alternative" long options (-like -this but also --like --this).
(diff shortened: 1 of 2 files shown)


enh-google:
add a test that would have caught the bug?

landley:
A fully populated test suite would test every short option and every long alias, and all the corner
cases for high and low ranges in each, feed non-numeric values into each numeric option, and so on. (Possibly the template shell script to do that could be generated from the option string, although the expected output wouldn't be unless you wanted 'exact match' of 'whatever it happens to produce' which isn't helpful. That by itself wouldn't have caught the above because it would have generated tests that matched the code, but then regression testing against a second implementation ala IETF procedure would have caught the discrepancy.)

But until/unless I get a chance to open that can of worms, I don't see much point in adding a single "do any longopts work at all" test to this command and not the others. I agree "we found a bug, we should add a test" is a good principle, but it's a drop in a bucket in this instance.

enh-google:
certainly hard to imagine me regressing this. (though i do often wish we at least had some checking that the help text and the actual options understood by the toy matched. python's argument parsing stuff is nice that way.)

PR #157, opened 2019-12-12
fix loopback device mount fail

wayling:
When we "losetup" success need mount loop device.
Found this issue on AndroidQ

Relate to commit 8a9484e8bfbfc8cd

### DIFF
--- a/toys/lsb/mount.c
+++ b/toys/lsb/mount.c
@@ -270,6 +270,7 @@ static void mount_filesystem(char *dev, char *dir, char *type,
       dev = tortoise(1, (char *[]){"losetup",
         (flags&MS_RDONLY) ? "-fsr" : "-fs", dev, 0});
       if (!dev) break;
+      continue;
     }
 
     free(buf);


landley:
Does this work instead?

diff --git a/toys/lsb/mount.c b/toys/lsb/mount.c
index 296d686..cc13c1d 100644
--- a/toys/lsb/mount.c
+++ b/toys/lsb/mount.c
@@ -275,7 +277,7 @@ static void mount_filesystem(char *dev, char *dir, char *type,
     free(buf);
     buf = 0;
     if (!rc) break;
-    if (fp && (errno == EINVAL || errno == EBUSY)) continue;
+    if (fp && (!errno || errno == EINVAL || errno == EBUSY)) continue;

     perror_msg("'%s'->'%s'", dev, dir);

landley:
Sigh. Github does not support cutting and pasting a patch into comments.

It just adds !errno to the list of errno values that loop, but moves it down after freeing buf. Does that fix the problem for you? (You didn't give me a reproduction sequence...)

enh-google:
you need to put ``` before it and ``` after it. regular markdown blockquote syntax. you can edit your comment after the fact to fix it too.

landley:
Ah, I didn't know github has gaslighting support. (Can't say I'm surprised though.)

How do I "view raw" your comment so I can see what tags you used, since they got converted?

enh-google:
lol. that'll teach me not to try the preview when explaining markup... i don't know how to view raw for the _past_. so here it is again, in its own ``` block:

```
you need to put ``` before it and ``` after it.
```

wayling:
> Does this work instead?
> 
> diff --git a/toys/lsb/mount.c b/toys/lsb/mount.c
> index 296d686..cc13c1d 100644
> --- a/toys/lsb/mount.c
> +++ b/toys/lsb/mount.c
> @@ -275,7 +277,7 @@ static void mount_filesystem(char *dev, char *dir, char *type,
> free(buf);
> buf = 0;
> if (!rc) break;
> 
>  if (fp && (errno == EINVAL || errno == EBUSY)) continue;
> 
>  if (fp && (!errno || errno == EINVAL || errno == EBUSY)) continue;
>   perror_msg("'%s'->'%s'", dev, dir);

HI ,
I think not work because errno will keep 15(Block device required) on AndroidQ.
image

landley:
You're right, Section 7.5.3 in c99: initialized to 0 but never set _back_ to zero. (I used a broken solaris compiler in college that didn't do that, I've always just used or snapshot the value right after the call that set it ever since.)

Moreover I think your original fix is right, _not_ freeing/zeroing buf keeps type for the retry of mount() (avoids looking it up again) and then it gets freed normally on loop exit.

PR #171, opened 2020-02-26
lib/env: avoid use-after-free in xclearenv

pallas:
In `xclearenv`, `char **environ` is either NULL or it is freed.  Afterward,
it is dereferenced.  It appears that the intention was to NULL out the
variable, not write a NULL to the now-dangling array.

Signed-off-by: Derrick Pallas <derrick@pallas.us>

### DIFF
--- a/lib/env.c
+++ b/lib/env.c
@@ -29,7 +29,7 @@ void xclearenv(void)
     free(environ);
   }
   toys.envc = 0;
-  *environ = 0;
+  environ = 0;
 }
 
 // Frees entries we set earlier. Use with libc getenv but not setenv/putenv.


landley:
Thanks for the bug report. It needs a different fix though.

The real bug is it shouldn't free(environ) at all here: environ is never null and lots of stuff reaches out and dereferences it without checking for null (including the functions before and after this one).

At the design level, we don't modify our inherited environ[] the same way we don't modify our inherited argv[], so the else path should xmalloc() a new environ at the same stride as xsetmyenv() uses (256 entries) and then set toys.envc to 1 to indicate we're using our own.

See 3b82917a49f9

PR #177, opened 2020-03-26
du: Implement -b

chirayudesai:
This comes straight from http://lists.landley.net/htdig.cgi/toybox-landley.net/2019-December/011304.html

I needed it as a band-aid while building Android.

The Android build system calls toybox du to calculate some sizes [1]
I build on zfs compression enabled volume, which means that the current `du -k -s` does not return the correct size. Adding `-b` does get close to the actual size, enough that a quick smoke test build boots. cc @enh-google 

[1]: https://android.googlesource.com/platform/build/+/refs/tags/android-10.0.0_r31/tools/releasetools/build_image.py#60

### DIFF
--- a/toys/posix/du.c
+++ b/toys/posix/du.c
@@ -4,9 +4,11 @@
  *
  * See http://opengroup.org/onlinepubs/9699919799/utilities/du.html
  *
- * TODO: cleanup
+ * TODO: cleanup (should seen_inode be lib?)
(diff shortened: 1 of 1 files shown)


landley:
Applied, commit cc803fb5fb0a

chirayudesai:
Thanks for the super-fast merge!

AOSP patch for reference: https://android-review.googlesource.com/c/platform/build/+/1269603

PR #178, opened 2020-03-29
file: add 7z format

joeky888:
Magic number: '7', 'z', 0xBC, 0xAF, 0x27, 0x1C


landley:
Do you have a small example tar.7z file for the test suite?

joeky888:
I added a 7z file to tests.

landley:
Applied.

PR #179, opened 2020-04-15
allow mount -a -t notype,type2

E5ten:
type list starting with "no" no longer needs all types in list to have
"no" prefix


landley:
Test case please? (I know there isn't a mount.tests yet but could you post something here so I can have an example that works now and didn't before?)

E5ten:
Using filesystems that are relevant to my system, `mount -at noswap,vfat` gives "mount: bad typelist" before, and needed `mount -at noswap,novfat` to work, and now the former case works too.

PR #180, opened 2020-04-15
Fix various warnings

E5ten:


landley:
Applied patch 1 and 2, but patch 3 is a nightmare because fdisk.c is a nightmare and read_size() should probably just be a call to fdlength() and I'm not 100% certain the original submitters didn't do http://lists.landley.net/pipermail/toybox-landley.net/2015-April/004026.html there again and...

It hasn't made it out of pending for a reason. That said, I'm slowly switching the remaining uint64_t in the tree to just "unsigned long long" because LP64 and I want to be consistent. Plus that lets us keep printf("%llu") which is a lot simpler.

Anyway, I'm skipping #3 for now, FYI.

landley:
Fixed up one more patch, applied the other two.

PR #187, opened 2020-04-25
silence -Wstring-plus-int warnings throughout source, remove -Wno-string-plus-int

E5ten:
I get that this warning is dumb so I understand if it gets rejected, I just figured it might be preferable to not have to disable it (especially if it happens to eventually catch an actual issue).

### DIFF
--- a/lib/lib.c
+++ b/lib/lib.c
@@ -444,7 +444,7 @@ int unescape2(char **c, int echo)
   if (idx != '\\' || !**c) return idx;
   if (**c == 'c') return 31&*(++*c);
   for (i = 0; i<4; i++) {
-    if (sscanf(*c, (char *[]){"0%3o%n"+!echo, "x%2x%n", "u%4x%n", "U%6x%n"}[i],
+    if (sscanf(*c, (char *[]){&"0%3o%n"[!echo], "x%2x%n", "u%4x%n", "U%6x%n"}[i],
         &idx, &off))
     {
       *c += off;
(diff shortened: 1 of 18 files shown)


E5ten:
If the warning fixes are fine but the removal of probecc() isn't because there's potential for use in the future I can change that commit to just remove the -Wno-string-plus-int addition.

landley:
No, "string"+offset is valid C, the same way any other pointer+offset is well-defined C: it advances by offset*sizeof(*pointer) which in this case is 1 (LP64 says so and C99 section 6.5.3.4). If you add two strings together you should get an error (because pointer + pointer has no meaning). If you add two ints together you don't have a pointer and it warns about the implicit type conversion when you pass it as an argument.

What "actual issue" is this even theoretically supposed to catch? How is taking the address of an array offset supposed to be clearer? I can see if (assignment) instead of if (comparison) being a common enough typo to require the extra parentheses, but I don't understand what's supposed to be wrong with this one?

I'm not auditing these for what else they might have done, please resubmit the other changes separately.

E5ten:
I wasn't saying it wasn't valid, just wasn't sure if silencing the warning in the code instead of disabling it would be objected to, but I completely understand why. In regards to a theoretical issue, when I was looking for the origin of -Wno-string-plus-int in toybox I saw that enh said "as for the usefulness of this particular warning, we have seen it catch real bugs in OEM code :-(", not that it seems likely or even possible that said "real bug" would appear here. The only other changes I made were switching to FLAG() macros, should I make a new PR with those changes, and if so, should I go through the codebase and switch to FLAG() macros wherever relevant and include all of that in one big "switch to FLAG() macros" PR?

enh-google:
> when I was looking for the origin of -Wno-string-plus-int in toybox I saw that enh said "as for the usefulness of this particular warning, we have seen it catch real bugs in OEM code :-(", not that it seems likely or even possible that said "real bug" would appear here.

yeah, i think the only risk is if you're expecting a sudden influx of Java programmers trying to  make toybox changes. i'm pretty sure they'll fall at earlier hurdles than this :-)

landley:
I'm happy to write a toybox programming HOWTO with an entire "how to program in C" section at some point, but that's a post-1.0 activity. (Let's be honest, that's a world with Basic Income where I have time to DO that sort of thing activity.)

PR #189, opened 2020-04-27
stty.c: fix interpreting output of stty -g

### LINKED ISSUE #188: stty doesn't accept output from stty -g as input
```
stty "$(stty -g)"
stty: not integer: 500:5:bf:8a3b:3:1c:7f:15:4:0:1:0:11:13:1a:0:12:f:17:16:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0
```

E5ten:
move case checking for stty -g output as input above case checking for
an int argument, so atoi doesn't cause "500:5:bf:8a3b:3"... to be
considered an int argument of 500
Fixes #188


E5ten:
ping?

PR #190, opened 2020-04-29
Bootstrap toysh and sed without bash, make or sed.

nykula:
Ok: bash --posix, busybox sh, dash, ksh, mksh, pdksh, zsh --emulate sh.
Hopefully toysh, once it supports globs and case statements. Wontfix:
csh, fish, heirloom-sh, powershell, rc.

Doesn't resolve dependencies other than sh builtins. Still builds and
calls kconfig, config2help and mkflags. Translating my loops and trivial
pattern matching into more such C utilities could be the future key to
building in a non-posix shell like those listed above.

Non-goals: any optimization, asan, githash, isnewer, kconfig_config,
libprobe, menuconfig, mkdir-less build, optlibs, parallel, probesymbol,
running when generated/ exists, singlemake, strip, tagged_array, time.

### DIFF
--- a/scripts/mkba.sh
+++ b/scripts/mkba.sh
@@ -1,0 +1,93 @@
+#!/bin/sh
+# Bootstrap toysh and sed without bash, make, sed etc on host.
+# Required posix utilities: c99 mkdir printf sh test
+# 0BSD 2020 Denys Nykula <nykula@ukr.net>
+set -e
+CC="${CC:-c99 -D_DEFAULT_SOURCE -Wno-pedantic}"
+HOSTCC="${HOSTCC:-c99 -D_DEFAULT_SOURCE -Wno-pedantic}"
+export LANG=C LC_ALL=C depends="TOYBOX_HELP TOYBOX_HELP_DASHDASH" toyfiles=
+mkdir generated generated/sort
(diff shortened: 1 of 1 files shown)


landley:
Interesting. I'll have to look at it more closely after the release.

I note that "WRAPDIR=wrap wraplog=log.txt scripts/record-commands make distclean defconfig toybox" followed by "awk '{print $1}' log.txt | sort -u | xargs" says the full toybox build uses:

as basename cat cc chmod cmp dirname echo egrep find git grep head ld ln ls make mkdir nproc rm sed sort strip tail tee tr uname wc xargs

The "as cc ld make strip" binaries come from the toolchain, and git is optional (trying to --describe the version), which leaves "basename cat chmod cmp dirname echo egrep find grep head ln ls mkdir nproc rm sed sort tail tee tr uname wc xargs" (plus of course sh, which this doesn't catch because #!/blah isn't a $PATH lookup). So a make mini_airlock that builds just those binaries with minimal pure-posix host dependencies would be a very intersting script. :)

Rob

landley:
P.S. I've already established that toybox builds under itself. That's why scripts/mkroot.sh has an airlock step.

landley:
Try commit 3bbc31c78b41

PR #195, opened 2020-05-20
glibc on mips is missing SIGSTKFLT

kraj:
Do not therefore assume it being available linuxwide

Fixes
| lib/portability.c:433:3: error: use of undeclared identifier 'SIGSTKFLT'
|   SIGNIFY(STKFLT), SIGNIFY(POLL), SIGNIFY(PWR),
|   ^


enh-google:
mips doesn't have SIGSTKFLT. i've no idea why musl claims it's signal 7. the kernel disagrees, calling that SIGEMT :-)

landley:
Applied.

PR #196, opened 2020-05-21
Tackle SIGEMT and SIGSTKFLT is not glibc specific

kraj:


landley:
I already implemented my suggestion earlier today. Query: is SIGEMT also not there on mips? The earlier thread was just about SIGSTKFLT?

kraj:
Its only defined for mips/sparc/alpha in glibc so checking for definition is wise thing to do.

enh-google:
fix merged.

PR #197, opened 2020-05-26
Replace [gs]ettimeofday with clock_[gs]ettime

E5ten:
gettimeofday is XSI, not POSIX, and is marked obsolete, while settimeofday is non-standard. Switch use of these functions to their modern replacements.

### DIFF
--- a/toys/other/hwclock.c
+++ b/toys/other/hwclock.c
@@ -56,8 +56,7 @@ static int rtc_find(struct dirtree* node)
 
 void hwclock_main()
 {
-  struct timezone tzone;
-  struct timeval timeval;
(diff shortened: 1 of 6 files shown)


E5ten:
In addition to musl ignoring the settimeofday timezone argument, mentioned in the hwclock commit, glibc 2.31 and up apparently errors when settimeofday is called with both arguments non-NULL.

landley:
Six patches in one pull request.

E5ten:
Sorry about that, I'll split them up.

E5ten:
Made a new PR for each command changed, closing this one.

PR #198, opened 2020-05-26
devmem: replace getpagesize with POSIX sysconf(_SC_PAGESIZE)

E5ten:


landley:
applied

landley:
You know, I applied this (f29cde56e929) but I have a question: how is libc supposed to get this value? Isn't it just going to call the same syscall? Because the reason I _didn't_ hardwire it in as 4k is it varies per architecture, sometimes as a kernel config. (I don't know of any modern system where it's NOT 4k, and the only one with significant historical usage was apparently Alpha, but in theory the same libc can run with different page sizes. Meaning you need a system call or similar to find the info, and Linux HAS a system call, which is why I was calling it...)

How is "let somebody else make the exact same system call for you" a good thing? Gratuitous unnecessary wrapper pulled in out of a library that MUST do the same thing we're doing here? Yanking unnecessary layers of indirection is how toybox is small and simple.

What, specifically, is the proposed advantage?

enh-google:
(note that there are very few architectures where getpagesize() is actually a syscall --- stuff like alpha and sparc, nothing commercially relevant in 2020. instead there's an AT_PAGESZ in the ELF auxval stuff. but the general point probably still stands. interestingly, by historical accident bionic just returns the constant PAGE_SIZE for getpagesize() whereas sysconf() actually does the getauxval() lookup. funnily enough, most would-be callers seem to just inline PAGE_SIZE anyway, which makes it hard to know how we'd move up from 4KiB pages without breaking a lot of apps.)

PR #199, opened 2020-05-26
Remove XSI putenv usage, replace with POSIX setenv

E5ten:


enh-google:
(discussed on mailing list.)

PR #200, opened 2020-05-26
lib.[ch]: switch loggit to macro, to avoid non-standard vsyslog

E5ten:


E5ten:
Related question, is there a reason to keep loggit in lib, given that it's only used in su?

landley:
No.

landley:
I believe the original intent was to have the other users of syslog in pending go through that function? It's been a while. (But function to multiline macro expanding in each caller because posix hasn't caught up to reality is not a compelling change. What's it "portable" _to_?)

E5ten:
Fair enough, multiline macros are not ideal.

PR #201, opened 2020-05-26
dhcpd: fix warnings

E5ten:
Add presumably intended conversion specifier for gconfig.valid_lifetime
as it is an argument to error_msg despite the format string having no
conversion specifiers.
Remove always true false inverse check for address of array
gstate.rcvd.rcvd_pkt6.transaction_id.


landley:
Applied (3c01911b0f85) with %u instead of PRIu32 (we're LP64, but even on winders "int" is always 32 bits).

E5ten:
Awesome, I now get no warnings on an allyesconfig build, aside from the intended one in skeleton.c.

PR #203, opened 2020-05-27
Replace uses of obsolete usleep with POSIX nanosleep

E5ten:


landley:
There's msleep() in lib, used that instead.

PR #205, opened 2020-05-27
hwclock: switch obsolete non-POSIX [gs]ettimeofday to clock_[gs]ettime

E5ten:
Remove timezone handling, clock_settime does not support it, and the
timezone argument to settimeofday is a no-op on musl anyway.

EDIT: relevant comment I made on the previous PR that included this commit:
In addition to musl ignoring the settimeofday timezone argument, mentioned in the hwclock commit, glibc 2.31 and up apparently errors when settimeofday is called with both arguments non-NULL.

### DIFF
--- a/toys/other/hwclock.c
+++ b/toys/other/hwclock.c
@@ -56,8 +56,7 @@ static int rtc_find(struct dirtree* node)
 
 void hwclock_main()
 {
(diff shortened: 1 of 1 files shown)


enh-google:
(discussed on mailing list.)

PR #209, opened 2020-05-27
date: replace non-standard settimeofday with clock_settime

E5ten:


enh-google:
(discussed on mailing list.)

E5ten:
So I was looking at this again, regardless of the portability matter (clock_settime is POSIX since 1993 vs settimeofday which isn't), using settimeofday here also loses precision:
```c
    struct timeval tv;

    parse_date(setdate, &t);
    tv.tv_sec = t;
    tv.tv_usec = TT.nano/1000;
    if (settimeofday(&tv, NULL) < 0) perror_msg("cannot set date");
```
The time is in nanoseconds, but it's divided into microseconds for the timeval, while clock_settime and a timespec wouldn't lose precision. Based on that I think this change should be reconsidered.

E5ten:
"Use POSIX API" was my main reason anyway, if that's good enough for you then great lol

landley:
On 3/9/21 12:02 AM, E5ten wrote:
> "Use POSIX API" was my main reason anyway, if that's good enough for you then
> great lol

"Use posix api" is a reason to look into it, but not necessarily to do it. The
settimeofday man page says:

CONFORMING TO
       SVr4, 4.3BSD.  POSIX.1-2001 describes gettimeofday() but not settimeof‐
       day().  POSIX.1-2008 marks gettimeofday() as obsolete, recommending the
       use of clock_gettime(2) instead.

Which says to ME this is like ANSI C coming up with its own fopen() api because
it didn't like the existing unix file descriptors for some reason. (C and Unix
were created by the same people, an attempt to "standardize" by deciding you
don't LIKE some of the original plumbing for no obvious reason is not a reason
to move off the original plumbing?)

When you said "this change should be reconsidered" I thought you meant updating
a recent commit to date.c, but according to git annotate the settimeofday()
call's line in date.c is from 2012. I admit I'm not a fan of microseconds
because milliseconds and nanoseconds have more reason to exist (one's human
visible and the other's cpu visible):

  https://landley.net/notes-2018.html#06-10-2018

But in this case we can't achieve more precision than we're already doing and
the change seems like pointless churn. It has literally no benefit except
compliance with a standard that intentionally decided not to include it for no
obvious reason, and was UNIVERSALLY ignored in this. (Is there a vaguely unixish
build environment settimeofday() ISN'T available in? It's in sparc, aix, qnx,
mbed, newlib/libgloss...)

Personally, I always find clock_settime() a bit of a headache because it wants
to know WHICH clock to set (posix mentions 2, the linux man page mentions 8),
and I REALLY don't want to have to care. The difference between the starbucks
menu and just ordering "a coffee" from a place that will give you "a coffee"...

Rob

E5ten:
Unless you have specific reason to use a different one, you can just assume CLOCK_REALTIME is what you want, it's equivalent to what [gs]ettimeofday use.

E5ten:
Also, this file already uses the timespec clock_* API with clock_gettime right above, imo it makes sense to be consistent here and use clock_settime as well.

landley:
So if I use both open() and fopen() in parts of the same program I should definitely remove all uses of file descriptors since that's not what C99 decided to standardize?

E5ten:
I don't think that comparison is very fair, stdio and file descriptors are different functionality, settimeofday/clock_settime are the same functionality, with one non-standard and the other standard for decades.

Separately, in response to `When you said "this change should be reconsidered"...` because I missed that part before, what I meant by "this change" was this PR, as in it should be reconsidered for inclusion. Sorry for the confusion.

PR #211, opened 2020-05-28
Replace obsolete and non-standard gethostbyname{,2} with getaddrinfo

E5ten:


enh-google:
(discussed on mailing list.)

PR #212, opened 2020-05-28
Replace non-standard setlinebuf with setvbuf

E5ten:


enh-google:
(discussed on mailing list.)

PR #213, opened 2020-05-28
Switch from obsolete utmp interface to utmpx

E5ten:
On linux, utmp isn't obsolete, and utmpx is the same as it, but on pretty much any other system utmp is obsolete and utmpx should be used instead.

### DIFF
--- a/toys/pending/getty.c
+++ b/toys/pending/getty.c
@@ -27,7 +27,7 @@ config GETTY
 */
 #define FOR_getty
 #include "toys.h"
-#include <utmp.h>
+#include <utmpx.h>
 
 GLOBALS(
   char *issue_str;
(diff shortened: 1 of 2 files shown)


landley:
Android NDK 25b has utmp.h but not utmpx.h. Switching API would break building these on android.

enh-google:
> Android NDK 25b has utmp.h but not utmpx.h. Switching API would break building these on android.

yeah, while i'm always skeptical of "obsolete", and even more skeptical of anything being removed, a more persuasive version of this would have been "Switch from legacy utmp interface to POSIX utmpx". and since these are all no-ops on Android anyway, I can actually fix this for you in a header-file-only backwards-compatible way. amusingly, portability.h is _already_ using `__has_include()` so by adding the header locally, i've already tested toybox that way :-)

https://android-review.googlesource.com/c/platform/bionic/+/2521475

(if you want to make this change, you'll still need a newer NDK, of course. but toybox binaries built by it will run on any existing version of Android.)

PR #214, opened 2020-05-29
Remove semicolon following GLOBALS macro in the few commands that had one, as well as unneeded semicolons following one line cp functions

E5ten:
These cause a -Wextra-semi (if it is enabled) warning for every inclusion of globals.h, due to a double-semicolon following the structs for these commands' global variables. Remove the unnecessary semicolons.

### DIFF
--- a/toys/net/netstat.c
+++ b/toys/net/netstat.c
@@ -34,7 +34,7 @@ config NETSTAT
 GLOBALS(
   struct num_cache *inodes;
   int wpad;
-);
+)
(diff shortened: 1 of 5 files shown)


landley:
Does this change actually accomplish anything?

enh-google:
...it would mean i have no chance of winning the pedantry award for 0.8.4 :-)

landley:
Commit f9acf3b2f179

PR #215, opened 2020-05-29
replace non-standard zero-size arrays with flexible array members

E5ten:


landley:
Is there a reason? What compiler does toybox build with after this change that it didn't build with before?

PR #216, opened 2020-05-29
Remove GNU-extensions in variadic macros

E5ten:


landley:
The kconfig subdirectory is code copied from the linux kernel, which should be completely replaced (for license reasons as much as anything else, also because the current kernel version is turing-complete so "make oldconfig" can "rm -rf /home" and this one isn't). Random deltas from the upstream is pointless churn.

The other command is in pending, I highly doubt the macro will survive cleanup. A debug printf macro isn't the sort of thing that ships in a completed command.

PR #217, opened 2020-05-29
Remove use of GNU C extensions typeof, and statement expressions

E5ten:


landley:
toybox's maxof() and minof() don't have double evaluation problems, they're defined in lib/lib.h:

#define minof(a, b) ({typeof(a) aa = (a); typeof(b) bb = (b); aa<bb ? aa : bb;})
#define maxof(a, b) ({typeof(a) aa = (a); typeof(b) bb = (b); aa>bb ? aa : bb;})

It snapshots them into local variables with the same type as the argument precisely so it doesn't evaluate the argument twice.

So first you remove a use that's valid for the existing implementation, and then you break the implementation. What on earth is the point of this? This works fine in llvm, what broken compiler are you trying to work around the deficiencies of? Or is this "infrastructure in search of a user" where you think someday someone might want to build on Turbo C?

Are you going to remove the ~2000 uses of typeof() from the linux kernel? Or do you expect to build with a compiler that doesn't build the linux kernel? (If so, which compiler?)

~/linux$ toybox grep -rwM '*.[ch]' typeof | wc -l
1919

PR #218, opened 2020-05-29
Remove anonymous unions

E5ten:
These are either GNU extensions to c99/c11, the toybox design page says it's supposed to be c99, so name previously anonymous unions "u".

### DIFF
--- a/toys/example/skeleton.c
+++ b/toys/example/skeleton.c
@@ -55,7 +55,7 @@ GLOBALS(
     struct {
       long b;
     } a;
-  };
+  } u;
 
   int more_globals;
 )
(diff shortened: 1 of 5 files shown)


landley:
Read the comment at https://github.com/landley/toybox/blob/114541b98/toys/pending/sh.c#L1134 to see "yes, I am aware of this, I made an explicit choice to do it this way" because otherwise every reference to this gets gratuitously longer for no reason. It bloats the code for no benefit.

Portability TO WHAT? FOR WHO? What specifically are you trying to accomplish here? This builds with gcc and with llvm and presumably with intel's icc and I used to maintain a tinycc fork that I can pick up again and make compile this. The required behavior is not ambiguous.

landley:
The design.html page mentions c99 twice. The first says "we depend on c99 and posix-2008 features" but then immediately talks about how we depend on stuff that isn't in either. (I.E. we're using those as a baseline. Necessary but not sufficient.) The second mention of c99 says "c99 allows this but don't do it". The third mention just says we use // style comments more than /* */ because you never have to worry about terminating a //, the issue of "nesting" doesn't come up ether, and I have yet to personally encounter an asterisk-at-start-of-line alignment war over multiline //.

It does not say we never use anything that ISN'T in a standard.

E5ten:
Sorry, I took the paragraph talking about c99 (the one following "I don't do windows.") to mean that library types/functions/headers, and things related to the environment it's used in (like /proc) that aren't part of c99 were also required, I didn't realize that actual language extensions on-top of c99/posix were also included in the kind of extra stuff that was required.

PR #220, opened 2020-05-29
void pointer arithmetic is a GNU extension, get rid of it.

E5ten:


enh-google:
(discussed on mailing list.)

PR #222, opened 2020-05-30
void functions shouldn't return void expressions, separate the expressions from the return calls.

E5ten:


landley:
Replacing 1 line with 4 lines that do the exact same thing is an improvement how? (I did this intentionally. What you're saying is that you disagree.)

I agree the return in df.c is unnecessary.

landley:
Removed the one nonfunctional return in aafe864873bf

PR #223, opened 2020-05-30
2 miscellaneous C99 conformance fixes.

E5ten:


enh-google:
i think these were rejected on the mailing list as "gcc and clang don't care, so not a problem".

PR #236, opened 2020-08-17
toybox: include more web links in help text.

### LINKED ISSUE #50: When I call any applet with the --help option, please show this general information first
Hi! When I call any applet with the `--help` option, it would be good if you could please start the output with the following lines, then a newline.

> Toybox multi-call binary `$(git describe)`
> Web page: http://landley.net/toybox/
> Bug reports: https://github.com/landley/toybox/issues
> Feedback: http://lists.landley.net/listinfo.cgi/toybox-landley.net
> Contributors, see: https://landley.net/toybox/cleanup.html#intro

Some users have Toybox because it came preinstalled on their device. The above information will make it easier for such users to figure out where and how to submit bug reports and patches.

`git describe` looks at tag names and commits, and hopefully outputs a version number like "0.7.1-106-g409a8e0" or "0.7.2".

unforgettableid:
This commit partially fixes #50. For a more-complete fix, we'd also need to prepend at least one or two web links to the help for each individual applet.

### DIFF
--- a/Config.in
+++ b/Config.in
@@ -16,6 +16,9 @@ config TOYBOX
 	  usage: toybox [--long | --help | --version | [COMMAND] [ARGUMENTS...]]
 
 	  Toybox multicall binary: https://landley.net/toybox/about.html
+	  Bug reports: https://github.com/landley/toybox/issues
+	  Feedback: http://lists.landley.net/listinfo.cgi/toybox-landley.net
+	  Contributors, please see: https://landley.net/toybox/cleanup.html
 
 	  With no arguments, "toybox" shows available COMMAND names. Add --long
 	  to include suggested install path for each command, see


landley:
No. I don't want to do that. The longer the page is, the less likely people are to read it.

1) Given one URL to a site with a nav bar, I'm assuming they can find more there. It's identified the project, which was the objection. (If you want that nav bar to have a "reporting bugs" link, I can add that. Probably belongs in the FAQ anyway...)
2) I prefer bug reports on the mailing list, here is second choice (and android's bug reporter is third).
3) That page is a read-only archive you can't post through. Even if I thought the mailing list was relevant to put directly into the help text (I don't, it's linked from the page I already point them at), that wouldn't be how.

PR #237, opened 2020-08-19
Makefile tweaks, including fixing a typo.

unforgettableid:


landley:
The middle hunk is a legitimate fix, the other two are not.

landley:
Commit df71fc33af72

PR #240, opened 2020-09-09
Fix typo

firasuke:
Just a small typo.

Thanks for your efforts in creating this detailed list.


enh-google:
this actually isn't a typo: `[` really is a command :-)

see https://www.gnu.org/software/coreutils/manual/coreutils.html#test-invocation for more.

firasuke:
Oh, I can't believe I forgot about `[` xD

But given the provided notation, I quickly assumed that either this was added by mistake or there is a missing closing square bracket.

Thanks for clarifying!

PR #242, opened 2020-09-19
cksum and hostid are already implemented

firasuke:
Both of [cksum](https://github.com/landley/toybox/blob/master/toys/posix/cksum.c) and [hostid](https://github.com/landley/toybox/blob/master/toys/example/hostid.c) are already implemented, unless you're planning on removing them.


landley:
hostid was removed from defconfig (moved to the "examples" directory) because it's useless:
http://lists.landley.net/pipermail/toybox-landley.net/2016-December/008775.html

enh-google:
merged

PR #244, opened 2020-09-19
Replace $CC with cc

firasuke:
Having both `$CROSS_COMPILE` and `$CC` will cause a duplication of the target tuple (triplet or quadruplet).

This is mostly due to the fact that many cross compilation toolchains set `CROSS_COMPILE` to `$TARGET_TRIPLET-` and `CC` to `$TARGET_TRIPLET-cc`, so in the case for `scripts/portability.sh` the final path will be `$TARGET_TRIPLET-$TARGET_TRIPLET-cc` which won't be found.

### DIFF
--- a/scripts/portability.sh
+++ b/scripts/portability.sh
@@ -2,9 +2,9 @@
 
 source configure
 
-if [ -z "$(command -v "${CROSS_COMPILE}${CC}")" ]
+if [ -z "$(command -v "${CROSS_COMPILE}cc")" ]
 then
-  echo "No ${CROSS_COMPILE}${CC} found" >&2
+  echo "No ${CROSS_COMPILE}cc found" >&2
   exit 1
 fi
 


landley:
Except it's not always cc, the compiler name is sometimes "android29-clang".

firasuke:
Yes, I understand that, but that's not the point. Most toolchains assume that the CC variable is already prefixed by the target tuple. To get this to work on these toolchains, one will either have to set CC equal to the compiler's name without the tuple (be it gcc or clang or cproc or whatever) or leave CROSS_COMPILE empty and prefix CC with the target tuple.

Using your example, say I'm cross compiling to `android29` so I set the variable `CROSS_COMPILE` equal to `android29-` and then I have `CC` set to `android29-clang` the resulting path will be `android29-android29-clang` which is wrong.

I also see that you're using ${CROSS_COMPILE}cc in mkroot.sh.

samboy:
My understanding of these conventions is that:

* `$CC` is the compiler which compiles to the target machine, and, if cross-compiling, is the cross compiler
* `$HOSTCC` is the compiler which makes binaries for the compiling machine, for cases when C programs have to be compiled locally.

Using `cc` instead of `$CC` is bad practice.

firasuke:
Your understanding is correct. But having both set will cause a duplication in the resulting value, so instead of having `x86_64-pc-linux-musl-gcc` as the cross compiler, it will evaluate to `x86_64-pc-linux-musl-x86_64-pc-linux-musl-gcc` which doesn't exist...

landley:
Supplying the prefix twice means it will be applied twice, and supplying the prefix in both places is supplying the prefix twice.

Personally I find it hard to get too worked up about this when the kernel guys ignored https://lkml.iu.edu/hypermail/linux/kernel/2202.0/01505.html because there's not much of a "toolchains do this standard thing" argument with the kernel special-casing its own "I am using LLVM" variable. Easily fixable (as demonstrated by the patch), they just don't seem to care...

landley:
Back in january, commit 6581ceaad5de updated the FAQ to clarify that supplying the prefix twice applies it twice.

firasuke:
@landley Thanks for updating the FAQ.

PR #245, opened 2020-10-05
Update i2ctools.c

chrisrfq:
Fix i2cdetect parameter reading so "last" value is read from correct argument.

Both range variables, "first" and "last" are incorrectly being read from argument 1. "last" should be read from argument 2.
This patch fixes the error.


enh-google:
merged as a24efe483b587f2af13fad29de6580fdd64a8407

enh-google:
(oh, and "thanks for fixing my mistake" :-) )

PR #249, opened 2020-10-13
Add some option alias

frikiluser:
Added some alias for existent options in various toys.

These alias are used by some building scripts for different tools considered "key-packages".

### DIFF
--- a/toys/lsb/gzip.c
+++ b/toys/lsb/gzip.c
@@ -8,7 +8,7 @@
  * todo: qtv --rsyncable
 
 // gzip.net version allows all options for all commands.
-USE_GZIP(NEWTOY(gzip,    "ncdfk123456789[-123456789]", TOYFLAG_USR|TOYFLAG_BIN))
+USE_GZIP(NEWTOY(gzip,    "n(no-name)cdfk123456789[-123456789]", TOYFLAG_USR|TOYFLAG_BIN))
 USE_GUNZIP(NEWTOY(gunzip, "cdfk123456789[-123456789]", TOYFLAG_USR|TOYFLAG_BIN))
 USE_ZCAT(NEWTOY(zcat,     "cdfk123456789[-123456789]", TOYFLAG_USR|TOYFLAG_BIN))
 
--- a/toys/posix/cp.c
+++ b/toys/posix/cp.c
@@ -16,8 +16,8 @@
 // for FLAG macros to work out right in shared infrastructure.
 
 USE_CP(NEWTOY(cp, "<2(preserve):;D(parents)RHLPprdaslvnF(remove-destination)fiT[-HLPd][-ni]", TOYFLAG_BIN))
-USE_MV(NEWTOY(mv, "<2vnF(remove-destination)fiT[-ni]", TOYFLAG_BIN))
-USE_INSTALL(NEWTOY(install, "<1cdDpsvm:o:g:", TOYFLAG_USR|TOYFLAG_BIN))
+USE_MV(NEWTOY(mv, "<2v(verbose)nF(remove-destination)fiT[-ni]", TOYFLAG_BIN))
+USE_INSTALL(NEWTOY(install, "<1cdDp(preserve-timestamps)svm:o:g:", TOYFLAG_USR|TOYFLAG_BIN))
 
 config CP
   bool "cp"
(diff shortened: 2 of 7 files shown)


frikiluser:
I've found some alias aren't listed when _--help_. Can I open a new pull-request adding them?

In addition to these option alias I'm building a list of _missing_ options. Some of them are already documented in TODO comments. Is it preferable to open a github issue with the complete list or adding them all to toys TODO comments and open a pull-request?

Regards,

enh-google:
> I've found some alias aren't listed when _--help_. Can I open a new pull-request adding them?

as i understand rob's position, he doesn't like long options but admits that they exist. so they _work_ (when someone actually has a script that needs them), but he doesn't add them _until_ that user shows up, and he doesn't want them in the help text (unless they're long-*only* options that don't have a corresponding short option).

> In addition to these option alias I'm building a list of _missing_ options. Some of them are already documented in TODO comments. Is it preferable to open a github issue with the complete list or adding them all to toys TODO comments and open a pull-request?

do you have specific use cases that require them? that's definitely valuable. but if it's just "here's the diff of --help against some other implementation", that doesn't really add anything. (especially given the overall policy -- as i've internalized it from the examples i've seen -- of "add it when there's a user for it".)

frikiluser:
> > I've found some alias aren't listed when _--help_. Can I open a new pull-request adding them?
> 
> as i understand rob's position, he doesn't like long options but admits that they exist. so they _work_ (when someone actually has a script that needs them), but he doesn't add them _until_ that user shows up, and he doesn't want them in the help text (unless they're long-_only_ options that don't have a corresponding short option).

OK

> > In addition to these option alias I'm building a list of _missing_ options. Some of them are already documented in TODO comments. Is it preferable to open a github issue with the complete list or adding them all to toys TODO comments and open a pull-request?
> 
> do you have specific use cases that require them? that's definitely valuable. but if it's just "here's the diff of --help against some other implementation", that doesn't really add anything. (especially given the overall policy -- as i've internalized it from the examples i've seen -- of "add it when there's a user for it".)

Suggested long options for this pull request are found in "real world" scripts:
* https://codesearch.debian.net/search?q=gzip.*--no-name&literal=0
* https://codesearch.debian.net/search?q=mv.*--verbose&literal=0
* https://codesearch.debian.net/search?q=install.*--preserve-timestamps&literal=0
* https://codesearch.debian.net/search?q=date.*--utc&literal=0
* https://codesearch.debian.net/search?q=file.*--brief&literal=0&page=1
* https://codesearch.debian.net/search?q=grep.*--count&literal=0
* https://codesearch.debian.net/search?q=rm.*--force&literal=0
* https://codesearch.debian.net/search?q=xargs.*--null&literal=0
* https://codesearch.debian.net/search?q=xargs.*--no-run-if-empty&literal=0

I'm building a list of _missing_ options the same way. I also have some notes about how frequent are used in this context.
Example _cp --reflink_:
* https://codesearch.debian.net/search?q=cp.*--reflink&literal=0&page=1

The good news are the existing options work pretty well :ok_hand:

landley:
It's not that I actively keep long options _out_ of the help text: hwclock and nsenter have longopt synonyms at the end of  lines, losetup has --show as an alias for ls, and when it's the ONLY way to do something it goes at the end ala ls. And for things like blkdev it just lists them.

But it's a judgement call how much toybox help should say at the best of times. I want to briefly explain how to use the command. What does that mean exactly? All the shrugging.

There IS a semi-policy of having undocumented options where we didn't actually implement something but we accept the option for compatibility reasons with a good enough NOP behavior. I just added the requested uname -p but didn't add it to the help text because it only ever says "unknown".  There's a bunch of these: patch has --no-backup-if-mismatch but it doesn't create backups so the option is telling it to do what it's already always doing...

Not mentioning a supported longopt as a synonym for a short opt isn't so much policy as "if they ask how to do X with this tool the answer is available" and calling it good enough, especially if adding the --longopt at the end would cause an 80 char line wrap when all the other option descriptions managed to fit in one line. "Is this clarifying or bloat? Does it let them do more than they could before?" It's aesthetic handwaving not a strongly held belief, and if somebody cares way more about it than I do...

landley:
Try commit 19f16d98258b

PR #253, opened 2020-12-05
id.test: hacks to pass on Raspberry Pi OS.

enh:
The rest of `make tests` all pass on a Raspberry Pi 400.

### DIFF
--- a/lib/lib.c
+++ b/lib/lib.c
@@ -349,17 +349,18 @@ int stridx(char *haystack, char needle)
 // Convert wc to utf8, returning bytes written. Does not null terminate.
 int wctoutf8(char *s, unsigned wc)
 {
-  int len = (wc>0x7ff)+(wc>0xffff), mask = 12+len+!!len;
+  int len = (wc>0x7ff)+(wc>0xffff), i;
 
   if (wc<128) {
     *s = wc;
     return 1;
   } else {
+    i = len;
     do {
-      s[1+len] = 0x80+(wc&0x3f);
-      wc >>= 7;
-    } while (len--);
-    *s = wc|mask;
+      s[1+i] = 0x80+(wc&0x3f);
+      wc >>= 6;
+    } while (i--);
+    *s = (((signed char) 0x80) >> (len+1)) | wc;
   }
 
   return 2+len;
--- a/tests/id.test
+++ b/tests/id.test
@@ -5,13 +5,15 @@
 #testing "name" "command" "result" "infile" "stdin"
 
 # Systems with SELinux will have security context cruft,
-# and BSDs call the root group "wheel" instead.
-CLEAN="sed 's/ context=.*//g' | sed 's/wheel/root/g'"
+# BSDs call the root group "wheel" instead,
+# and Raspberry Pi OS has root also in the 117(lpadmin) group.
+CLEAN="sed 's/ context=.*//g' | sed 's/wheel/root/g' | \
+sed 's/117//g' | sed -E 's/\(?lpadmin\)?//g' | sed 's/[ ,]$//'"
 
 testing "0" "id 0 | $CLEAN" "uid=0(root) gid=0(root) groups=0(root)\n" "" ""
 testing "root" "id root | $CLEAN" \
   "uid=0(root) gid=0(root) groups=0(root)\n" "" ""
-testing "-G root" "id -G root" "0\n" "" ""
+testing "-G root" "id -G root | $CLEAN" "0\n" "" ""
 testing "-nG root" "id -nG root | $CLEAN" "root\n" "" ""
 testing "-g root" "id -g root" "0\n" "" ""
 testing "-ng root" "id -ng root | $CLEAN" "root\n" "" ""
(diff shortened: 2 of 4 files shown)


landley:
Looks mostly good, what's the behavior change to lib.c? (What's the code change trying to accomplish?)

I note that "the max utf8 encoding is 4 bytes" is a hardwired assumption the unicode guys added a while back, which I was unhappy about at the time but have since used for buffer size allocations in a few places. See https://landley.net/notes-2017.html#01-09-2017

enh:
(sorry that i have no idea how to do pull requests. i wanted these to be two different pull requests, but obviously that didn't work out!)

> Looks mostly good, what's the behavior change to lib.c? (What's the code change trying to accomplish?)

the existing code produces wrong results, and it seems like it's otherwise only used in sh.c so far, which is why we haven't noticed.

specifically, each non-initial utf8 byte has _6_ meaningful bits not 7 [citation needed](https://en.wikipedia.org/wiki/UTF-8#Encoding), and the initial byte of any multibyte sequence was wrong. (i couldn't work out what you were trying to do with the additions, so rather than try to fix it i just replaced it with the obvious "smear the top bit by the right number of bits").

> I note that "the max utf8 encoding is 4 bytes" is a hardwired assumption the unicode guys added a while back, which I was unhappy about at the time but have since used for buffer size allocations in a few places. See https://landley.net/notes-2017.html#01-09-2017

yeah, i only added tests up to the first block of the first supplementary plane (i.e. U+10000) on the theory that we can worry about the rest if/when anyone cares. but since we already know what the maximum theoretical limit is (U+7fffffff, 6 bytes), we could just fix it now and never have to think about it again. i can send a patch for that too if you like.

enh:
(or i can send a patch to follow https://tools.ietf.org/html/rfc3629#section-3 and reject anything higher than U+10FFFF. that does make the API a bit more awkward though.)

landley:
Ah, new stuff for in-progress sh.c, not the old paths I (thought I) thoroughly tested in 2017. Got it.

I think most of the buf checks are just "refuse to overwrite the buffer if we're near the end and there isn't enough space", and hardwiring in "4" for that at the moment can't hurt anything, so my earlier unease was moot.

The UTF8 conspiracy declared that microsoft paid them to put an upper limit on the range forevermore, or some such. I had to source all this stuff to specs in 2017, which was long enough ago the details are blurring. But if you think supporting the full 6 bytes is useful, I'm happy to change stuff. I'm just trying to get the error return codes right. :)

Speaking of which, I point your attention to demo_utf8wc in the examples directory. (Used to be called test_, probably still should be called test_. I think it got renamed because make test_test_utf8towc looked ridiculous.) Oddly, it now has a delta with all three libc implementations, for reasons I am unclear on. Hmmm. (I built a historical version and it seems the libc implementations are what changed? Don't have brain to go down that rathole right now.)

landley:
Applied both.  I note that "unicode" and "ascii" are sort of analogous commands.

enh-google:
> I think most of the buf checks are just "refuse to overwrite the buffer if we're near the end and there isn't enough space", and hardwiring in "4" for that at the moment can't hurt anything, so my earlier unease was moot.

in wctoutf8, this is wrong for code points past the BMP (aka plane 0):
```
  int len = (wc>0x7ff)+(wc>0xffff), i;
```
utf8towc (which is obviously much more heavily used!) explicitly disallows anything past plane 1, so they should probably at least match?

> The UTF8 conspiracy declared that microsoft paid them to put an upper limit on the range forevermore, or some such. I had to source all this stuff to specs in 2017, which was long enough ago the details are blurring. But if you think supporting the full 6 bytes is useful, I'm happy to change stuff. I'm just trying to get the error return codes right. :)

yeah, i don't have an opinion either way on 4 vs 6, just that we should have wctoutf8 _either_ do the right thing _or_ return an error (rather than silently do the wrong thing, as now). i'm happy to send you a patch going in either direction, but was hoping you'd express a preference :-)

in practical terms, i think 4 is fine, and like you say, it matches MB_CUR_MAX.

> Speaking of which, I point your attention to demo_utf8wc in the examples directory. (Used to be called test_, probably still should be called test_. I think it got renamed because make test_test_utf8towc looked ridiculous.) Oddly, it now has a delta with all three libc implementations, for reasons I am unclear on. Hmmm. (I built a historical version and it seems the libc implementations are what changed? Don't have brain to go down that rathole right now.)

i think the "oddly..." part is the point --- if it's not run as part of `make tests`, it's not buying us anything :-(

> Applied both. I note that "unicode" and "ascii" are sort of analogous commands.

yeah, since `unicode` with no arguments isn't obviously useful (and does nothing on Plan9/Inferno) i did wonder about merging that functionality -- especially because i copy & pasted the handy list of names for the `<' '` control characters -- but i've always assumed that when man(1) is finished, ascii(1) gets removed in favor of `man ascii`?

also it just seemed weird to say "unicode" to get an ascii table :-) (though it would be weirder to say "ascii" to talk about unicode code points!)

happy to send a patch to make `unicode` behave like `ascii` if you want though.

PR #258, opened 2020-12-28
fix segfault in date

E5ten:
check if argument to -I is non-NULL before dereferencing

### DIFF
--- a/toys/posix/date.c
+++ b/toys/posix/date.c
@@ -138,7 +138,7 @@ void date_main(void)
 
   if (FLAG(I)) {
     char *iso_formats[] = {"%F","%FT%H%:z","%FT%R%:z","%FT%T%:z","%FT%T,%N%:z"};
-    int i = stridx("dhmsn", *TT.I ? *TT.I : 'd');
+    int i = stridx("dhmsn", (TT.I && *TT.I) ? *TT.I : 'd');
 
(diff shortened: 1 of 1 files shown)


landley:
The problem is actually that --iso sets this NULL but -I sets it to "" and that's inconsistent. The short opt should probably also set to NULL...

E5ten:
Ah I see, is that a problem in date specifically or just a general thing with lib/args.c and short+long opts that have optional arguments?

landley:
Two codepaths in lib/args.c plumbing handling "x(long):;" arguments two slightly different ways.

When I added ; to make an argument optional it was initially only used with bare longopts (ls --color I think), and leaving it NULL meant the flag had occurred without an argument. The short argument path is treating a lack of argment as a "" argument, ala -xabc == --long=abc but -x == --long= not --long

landley:
I wanted the fix to go the other way: --longopt= should return "" and --longopt should return NULL. That way you can distinguish "they did not provide an argument".

E5ten:
So is the longopt behaviour right now fine and just the shortopts need to be switched to return NULL instead of "", or is there longopt behaviour that also needs fixing?

enh-google:
looks like rob quietly fixed this with e7c86f7d79f641ab8a0e542d32e857494e943389...

PR #263, opened 2021-01-26
nc: always read all data if bigger than libbuf

asm89:
Previously nc would only read up to sizeof(libbuf) bytes in some cases.
For example, if a program writes more than 4kB of data to nc's stdin
_and_ closes stdin before nc has processed all the data, then nc would
read partial data. In this case what happens is that poll() returns both
POLLIN and POLLHUP for the fd. The code first handled the POLLIN and
then the POLLHUP in a single loop iteration. Instead it should handle
one event at the time, preferring POLLIN which will be set as long as
there is data available to read.

Improve bookkeeping of open fds in pollinate(). Clean up fds when read()
returns an error or EOF. Ensure we only clean up an fd once. Handle
POLLIN before POLLHUP.

To test that all data is now read (and sent) as expected. Start nc in
listening mode and have it reply with the number of bytes:

```
$ ./toybox nc -L -p 1234 wc -c
```

Create a test file and send it over to the listening side in various
ways. Confirm the correct byte count is returned each time:

```
$ python3 -c 'print("a" * 4096 + "b", end="")' > testfile
$ wc -c testfile
4097 testfile
$ ./toybox nc localhost 1234 < testfile
4097
$ cat testfile | ./toybox nc localhost 1234
4097
$ (cat testfile; sleep 1) | ./toybox nc localhost 1234
4097
```

### DIFF
--- a/lib/net.c
+++ b/lib/net.c
@@ -100,6 +100,7 @@ int pollinate(int in1, int in2, int out1, int out2, int timeout, int shutdown_ti
 {
   struct pollfd pollfds[2];
   int i, pollcount = 2;
+  int openfds = 2;
 
   memset(pollfds, 0, 2*sizeof(struct pollfd));
   pollfds[0].events = pollfds[1].events = POLLIN;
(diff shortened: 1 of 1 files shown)


asm89:
@landley This PR contained some other potential cleanups as well. Not sure if you're interested in those. I pulled out https://github.com/landley/toybox/pull/264 by itself. Happy to rebase the fd bookkeeping to be based on your `continue` commit as well.

landley:
I fixed the issue you sent me a test for.

PR #264, opened 2021-01-27
nc: stop polling closed fd

asm89:


landley:
There's no test to demonstrate this problem.

The pollcount-- means the next poll() is only of the first pollfds[] entry, and this is in an if (i) block so you're only ever setting the fd to -1 in the structure that won't be used again. What are you trying to accomplish here?

asm89:
Yes, you are right, `i` is only ever 0 or 1 here. This makes no sense without the other changes. Sorry!

PR #269, opened 2021-02-13
Add -executable & -executable-sane to find

aheirman:
I know the reasoning behind -executable also returning searchable directories. 
I just think it's bad and not intuitive for newcomers, `-executable-sane` should point out the default behavior is weird. 
The text next to `-executable` explains the behavior but a user that skims that over that might still see `-executable-sane`.

### DIFF
--- a/toys/posix/find.c
+++ b/toys/posix/find.c
@@ -34,6 +34,9 @@ config FIND
     -depth           ignore contents of dir    -maxdepth N at most N dirs down
     -inum N          inode number N            -empty      empty files and dirs
     -type [bcdflps]  type is (block, char, dir, file, symlink, pipe, socket)
+    -executable      Matches files which are executable and directories which
+                     are searchable
+    -executable-sane Matches files which are executable
     -true            always true               -false      always false
     -context PATTERN security context
     -newerXY FILE    X=acm time > FILE's Y=acm time (Y=t: FILE is literal time)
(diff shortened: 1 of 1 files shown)


landley:
Are you adding -executable speculatively, or is there an existing script that needs it? (It's not new functionality, you can "/perm a+x" already and your new proposal just adds "-type f".)

landley:
P.S. The suffix "-sane" you've invented is editorializing based on your opinion differing from everyone else who's contributed to the gnu or busybox find commands. It's a -type f constraint which already exists, and the option does not exist in the posix, bsd, solaris, or aix find commands.

I'm not merging that.

Given that -executable itself is there in busybox, I'm willing to be convinced that some existing script needed it... but git annotate says Denys added it in July 2018 in a commit saying nothing but that he added it (no reason why) and a quick check of the busybox mailing list web archive for july, june, and may of 2018 doesn't find any relevant mailing list messages with "find" or "executable" in the subject line. Looks like he was just filling out the man page options for the sake of doing so? (Implementing it because it was there, not because somebody needed it.)

The gnu man page says that the access() system call checks ACLs, which is a distinction you didn't mention...

aheirman:
> Are you adding -executable speculatively, or is there an existing script that needs it

I wanted to use it in a script on an OS that will never have GNU find.

> I'm not merging that.

I understand, would you be interested in merging this if `-executable-sane` is removed and the ACL bit is clarified?

aheirman:
I probably shouldn't have made a pr for a feature I don't like.

landley:
Commit 5c73305f0e4b

PR #273, opened 2021-03-11
df: Allow -P and -k options together as defined in POSIX

cpixl:
As defined in [POSIX Issue 7, 2008 Edition](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/utilities/df.html):

>When both the **-k** and **-P** options are specified, the following header line shall be written (in the POSIX locale):
>
>"Filesystem 1024-blocks Used Available Capacity Mounted on\n"
>
>When the **-P** option is specified without the **-k** option, the following header line shall be written (in the POSIX locale):
>
>"Filesystem 512-blocks Used Available Capacity Mounted on\n"

### DIFF
--- a/toys/posix/df.c
+++ b/toys/posix/df.c
@@ -4,7 +4,7 @@
  *
  * See http://opengroup.org/onlinepubs/9699919799/utilities/df.html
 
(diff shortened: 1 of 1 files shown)


landley:
commit 8c2825c496ae

PR #274, opened 2021-03-19
Handle paths starting with slash in rmdir -p

nir9:
When comparing behavior with the rmdir shipped with the linux distro I am running I have found that ```rmdir -p /a/b/c``` acts differently compared ```./toybox rmdir /a/b/c``` because of the way toybox handles the slashes, it removes them so when arriving at the first slash it actually attempts to rmdir nothing which is invalid.

This PR makes it correctly try to remove the first slash upon reaching there.

### DIFF
--- a/toys/posix/rmdir.c
+++ b/toys/posix/rmdir.c
@@ -37,6 +37,12 @@ static void do_rmdir(char *name)
     if (!toys.optflags) return;
     do {
       if (!(temp = strrchr(name, '/'))) return;
+
+      if (temp == name) {
+        temp[1] = 0;
+        break;
+      }
+
       *temp = 0;
     } while (!temp[1]);
   }


landley:
How about just:

--- a/toys/posix/rmdir.c
+++ b/toys/posix/rmdir.c
@@ -25,7 +25,7 @@ static void do_rmdir(char *name)
 {
   char *temp;
 
-  for (;;) {
+  do {
     if (rmdir(name)) {
       if (!FLAG(ignore_fail_on_non_empty) || errno != ENOTEMPTY)
         perror_msg_raw(name);
@@ -39,7 +39,7 @@ static void do_rmdir(char *name)
       if (!(temp = strrchr(name, '/'))) return;
       *temp = 0;
     } while (!temp[1]);
-  }
+  } while (*name);
 }

landley:
Wow, github hasn't got the concept of cutting and pasting a patch into a comment, does it? Of course wiki markup for comments is more important than C, why wouldn't it be? Sigh...

The patch changes the for loop into a do { ) while (*name);

enh-google:
```
Verbatim text here
```

nir9:
Hi,

The patch solves the problem of calling rmdir with nothing, but does not attempt to remove "/" like the rmdir I have compared to does.

Now I can agree that there isn't much point in attempting to remove "/", what do you think?

landley:
I specifically didn't want to remove / (at least not as part of a -p chain) since that doesn't make sense in unix.

Commit 855eca3eaf8e

PR #275, opened 2021-03-24
Fix install -t

### LINKED ISSUE #270: install -D -t <dir> does not create <dir>
This also doesn't work with `install -D file1 file2 dir`, where dir doesn't exist, but busybox also messes that up (differently, instead of making dir and putting file[12] there, it makes dir's parents and makes puts file2 as dir) so I'm not sure what should be done in that situation.
Either way, with any install I've seen that has `-t`, it creates the dir given to `-t` when `-D` is also passed.

E5ten:
I ran the tests with this and saw no regressions, and it seems to fix #270

### DIFF
--- a/tests/install.test
+++ b/tests/install.test
@@ -13,3 +13,7 @@ testing "install -D exists" \
   "mkdir -p a; touch a/b; install -D random a/b && cmp random a/b && echo yes" \
   "yes\n" "" ""
 rm -rf a random
+testing "install -D -t creates directory" \
+  "touch a; install -Dt b a && echo yes" \
+  "yes\n" "" ""
+rm -rf a b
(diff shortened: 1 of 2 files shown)


landley:
Could you add a tests/install.test entry demonstrating what it fixes?

(Due to the name conflict with "make install" there's no make test_install target, you have to run "scripts/test.sh install" directly. It's run as part of "make tests" though.)

E5ten:
Done, let me know if there's anything wrong with the test case, I think that's the first one I've written for toybox.

E5ten:
Also, in regards to the name conflict, I think the usual strategy here is to use xinstall throughout the source (like for the test name, the make target, etc.) and then to just install xinstall as install for `make install`. No idea how plausible or desirable it is to do something like this for toybox, but figured I'd bring it up in case it's a workable solution for the inability to run `test_install` individually.

E5ten:
Closed by https://github.com/landley/toybox/commit/3cde24138ac06e384965db2944e858d2110ecacb

PR #280, opened 2021-05-01
Add OpenBSD support

Ella-0:
Adds OpenBSD support. `chrt` and `fallocate` don't build due to missing functions.


landley:
Applied, but I note that the kconfig/ directory has freebsd_miniconfig and macos_miniconfig both of which are callable from kconfig/Makefile (and mentioned in the kconfig/Makefile help: target's output).

PR #284, opened 2021-05-19
Fix missing header file

### LINKED ISSUE #283: Cant Build passwd
when make:
```
toys/lsb/passwd.c: In function 'passwd_main':
toys/lsb/passwd.c:73:45: warning: implicit declaration of function 'getspnam'; did you mean 'getpwnam'? [-Wimplicit-function-declaration]
   73 |   if (*(pass = pw->pw_passwd)=='x' && (sp = getspnam(name))) pass = sp->sp_pwdp;
      |                                             ^~~~~~~~
      |                                             getpwnam
toys/lsb/passwd.c:73:43: warning: assignment to 'struct spwd *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
   73 |   if (*(pass = pw->pw_passwd)=='x' && (sp = getspnam(name))) pass = sp->sp_pwdp;
      |                                           ^
toys/lsb/passwd.c:73:71: error: invalid use of undefined type 'struct spwd'
   73 |   if (*(pass = pw->pw_passwd)=='x' && (sp = getspnam(name))) pass = sp->sp_pwdp;
```

it is not picking up getspnam

Ocean-Moist:
shadow.h is neccesary for ``getspnam``` and is missing

fix #283


landley:
That would break bionic, which doesn't have shadow.h.

The existing include of shadow.h in https://github.com/landley/toybox/blob/0.8.4/lib/portability.h#L215 is guarded by CFG_TOYBOX_SHADOW which is written into generated/Config.probed by https://github.com/landley/toybox/blob/0.8.4/scripts/genconfig.sh#L69 so the question is A) why is your build environment failing that compile-time probe, B) even if that IS failing why is the TOYBOX_SHADOW dependency in https://github.com/landley/toybox/blob/0.8.4/toys/lsb/passwd.c#L13 not preventing you from selecting the passwd command in your config?

PR #285, opened 2021-06-12
www: remove executable flag from html files.

ericonr:


landley:
http://www.html-faq.com/ssi/?xbithack

PR #288, opened 2021-06-27
Add support for -n in split

### LINKED ISSUE #287: (not posix, but gnu) split -n option
The `split -n` option is sometimes useful but sadly not present in the posix specification(https://pubs.opengroup.org/onlinepubs/009696699/utilities/split.html), but is is in the gnu version https://man7.org/linux/man-pages/man1/split.1.html .

I don't know of any scripts that use it, but it's handy.

Ella-0:
Fixes #287


landley:
Good start, but len is number of bytes read into a 4k buffer, not total file size.

Ella-0:
Fixed that, small problem; I'm unsure if using `lseek` is the best way to check if the value of `st.st_size` is valid.

Ella-0:
After investigating further it seams that this should only work for regular files.

landley:
Commits df7bfd2e1e79 and 3fbacb1f5c5e

landley:
Android builds with mksh? Could you provide more context please?

Ella-0:
Oh lol I pushed to the wrong branch. I meant to close this pr. As for building with mksh? wait -n is not implemented yet in mksh (though I'm working on patches) neither is the `<()` syntax.

Ella-0:
I think android gets away with it by generating config files ahead of time. See [here](https://android.googlesource.com/platform/external/toybox/+/refs/heads/master/android/linux/generated).

enh-google:
> I think android gets away with it by generating config files ahead of time. See [here](https://android.googlesource.com/platform/external/toybox/+/refs/heads/master/android/linux/generated).

yeah, we *test* with mksh (on the device), but we generate all the generated files ahead of time and check them in --- that's not part of the build at all.

the bashisms in the *build* are actually most annoying to me when fixing the mac build, because it makes the build a lot slower by falling back to the not-parallel path (as i've mentioned on the list)... the compilation part of the build is very fast on a current mac, but the bit before that regressed a lot.

Ella-0:
Sounds about right. If anyone's interested, I spent the day adding `<(...)` support to mksh. Patch found [here](https://github.com/iglunix/iglunix/blob/main/base/mksh/procsub.patch). I'll look into sending it upstream tomorrow.

enh-google:
interesting... my small bash-ism mksh patches have been rejected, but `<()` is by far and away the most common bash-ism i see people try to use, so you might have more luck! (mine were admittedly much less convincing; stuff like `$((3**4))` :-) )

Ella-0:
In the future plans section on the mksh website, it said that `<()` was something they might want to implement.

enh-google:
oh, nice! let me know if you get that in to mksh, and i'll make sure i update AOSP...

Ella-0:
`mksh` returns status 1 for that not 2 which is why I removed the line.

landley:
Wheee, version skew. I can update the test to "true & wait -n;" and then take any error return as "not supported". (In bash wait -n returns 127 when there was no processes to wait for, hence caring about the number...)

PR #289, opened 2021-07-13
fix the problem of compilation failure on the 3.10 kernel

linux-gcc:
fix the problem of compilation failure on the 3.10 kernel.

### DIFF
--- a/toys/other/lsattr.c
+++ b/toys/other/lsattr.c
@@ -111,8 +111,8 @@ static struct ext2_attr {
   {"Extents",                       FS_EXTENT_FL,       'e'},
   {"No_COW",                        FS_NOCOW_FL,        'C'},
   {"Casefold",                      FS_CASEFOLD_FL,     'F'},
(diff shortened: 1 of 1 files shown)


linux-gcc:
@landley Please help review the code. thanks!

landley:
The 3.10 kernel shipped in 2013 and https://landley.net/toybox/faq.html#support_horizon is 7 years. The last LTS bugfix to that kernel (.108) was almost 4 years ago. Why are you trying to build under that kernel, and do you really need lsattr? (Can you switch it off in the .config instead?)

Hmmm... That said, the two symbols were added to the UAPI headers more recently, https://github.com/torvalds/linux/commit/68ce7bfcd995 added FS_INLINE in 2016. The proper fix would be to add #ifndef SYMBOL/#define/#endif stanzas to lib/portability.h with a comment about when they expire.

landley:
Except linux/fs.h isn't #included for portability.h to modify, blah. Alright, lemme take a stab at this...

landley:
Commit 281f30dde58c

PR #292, opened 2021-08-11
fix ip.c compilation failed

linux-gcc:
1、fix ip.c compilation failed 
2、commit 3642f3f

### DIFF
--- a/toys/other/lsattr.c
+++ b/toys/other/lsattr.c
@@ -111,8 +111,8 @@ static struct ext2_attr {
   {"Extents",                       FS_EXTENT_FL,       'e'},
   {"No_COW",                        FS_NOCOW_FL,        'C'},
   {"Casefold",                      FS_CASEFOLD_FL,     'F'},
-  {"Inline_Data",                   FS_INLINE_DATA_FL,  'N'},
-  {"Project_Hierarchy",             FS_PROJINHERIT_FL,  'P'},
+  {"Inline_Data",                   0x10000000,  'N'},
+  {"Project_Hierarchy",             0x20000000,  'P'},
   {"Verity",                        FS_VERITY_FL,       'V'},
   {NULL,                            0,                  0},
 };
--- a/toys/pending/ip.c
+++ b/toys/pending/ip.c
@@ -34,6 +34,7 @@ config IP
 #include <net/if_arp.h>
 #include <ifaddrs.h>
 #include <fnmatch.h>
+#include <netinet/ip.h>
 #include <linux/if_tunnel.h>
 
 #ifndef IP_DF


linux-gcc:
I want to use this patch to try to solve the problem of compiling errors on kernel-3.10. When opening the CONFIG_IP configuration item.
I don't know why this PR will also bring this patch 490d587. I don't want to submit this patch 490d587，I just want to submit this patch 3642f3f. 
@landley

landley:
What distro are you building it on, because it built fine for me? (Well, with -Wno-restrict anyway.) And the extra header shouldn't be a kernel version thing...?

linux-gcc:
I compiled it on centos 7.4. 
@landley

landley:
Ah, it is kernel version skew:

Compile ip................In file included from /usr/include/linux/if_tunnel.h:7,
                 from toys/pending/ip.c:38:
/usr/include/linux/ip.h:86:8: error: redefinition of 'struct iphdr'
 struct iphdr {
        ^~~~~

landley:
And removing the linux/if_tunnel.h include causes... 23 build errors.

landley:
I downloaded a centos 7.4 live cd and booted it under kvm, and the fix you need (for the problem that struct iphdr isn't defined and thus field iph is an incomplete type) is to #include <linux/ip.h> on the line before the #include <linux/if_tunnel.h>.

Which I would be willing to apply (with a comment what it's working around), except that the recent "wait -n" commit broke centos 7.4 because your bash is more than 7 years old (https://landley.net/toybox/faq.html#support_horizon).

That's also trivially fixable: just drop the -n from the two wait calls (then the build isn't parallel, but it completes). But this is a policy question: Centos has a 10 year time horizon and Toybox has 7. Accumulating fixes for old build environments past our stated time horizon seems... ?

Elliott: do you have an opinion here?

landley:
Commits f37e5bf24540 and 3cf5ec1e9568

enh-google:
> But this is a policy question: Centos has a 10 year time horizon and Toybox has 7. Accumulating fixes for old build environments past our stated time horizon seems... ?
> 
> Elliott: do you have an opinion here?

oh, man... CentOS... don't get me started :-)

CentOS is our "problem child" OS with the NDK too. 10 year old version of glibc + 10 year old kernel means that even if we're hermetic for things like curses (which of course we should be, and any time we're not is a bug we should fix), but we still have troubles just with the libc basics. this is one reason why we're actually looking at switching host builds from glibc to musl --- then we can statically link and just make this kind of thing go away. (we'd use a current musl in conjunction with the current bionic uapi headers, so our only worry is stuff like "do we cope with ENOSYS" or whatever. though i guess this conversation makes me realize i don't know whether musl has a 7 year or a 10 year cruft horizon...)

(i mean, i love that i can run current xubuntu and a current kernel on my 10 year old personal laptop, and i would be sad if that went away, and i also get LTS releases [and mostly prefer LTS releases], but running a 10 year old OS ... surely part of the point of linux is that i _don't_ have to do that! even by the end of a 3 year LTS period i'm starting to feel like "these shoes are too small, time for new shoes".)

luckily i don't know of anything _worse_ than CentOS, so at least 10 years seems like the absolute limit!

landley:
Supporting old hardware is great: I'm typing on a laptop introduced in 2012 (and I bought a couple spares for when I wear out this one). But that seems unrelated to running a newer build environment on that hardware? I upgraded from Devuan Anthill to Devuan Babypowder because i couldn't build qemu anymore because their git introduced a requirement for a python release newer than Lucifer's move to Netflix. That's ANOTHER project more or less driven these days by Red IBM, which has ZERO respect for anybody else's build environments.

I try to minimize toybox dependencies, and to build toybox on the centos kvm live cd I just had to "yum install gcc" and "yum install make". (Ok, I also installed git because I was lazy downloading it, and then "yum install ncurses-devel" to run menuconfig: rewriting that would remove both that dependency and clean the gplv2 build infrastructure out of the tree. But it didn't NEED those last two.)

I have a vague TODO item of someday (after I've finished toysh and implemented make) creating a scripts/oroborous.sh that would build a minimal toybox with the build tools needed to build toybox defconfig, so you could "PATH=$PWD/airlock:$PATH make distclean defconfig toybox" and go from there. (My own version of a hermetic build, only building from source instead of binaries, depending only on the compiler+linker and libc to be there plus very basic posix command line utilities. That first build would either use a snapshot "generated" dir or make a really simple one entirely via vanilla posix shell stuff, quite possibly mkdir + writing some HERE documents.) But alas, that's a ways off...

The toybox 7 year horizon thing is policy. You've gotta draw the line _somewhere_. I'm not sure it's the right policy, but Red Hat sets itself up to receive very large checks from organizations that want the luxury of running 10 year old configurations, and I got a new $5 patreon donor yesterday who I'm still kinda jazzed about and need to thank. Bit of a capability gap there.

I'm pretty good about supporting very old compilers. (I only moved off the last-gplv2 ones because peter anvin made building the kernel with them painful.). I have a 7 year time horizon for kernel system calls, which is generally pretty easy to track. (Did you know my integrated "full kernel history back to 0.0.1"  git repo got uploaded to kernel.org? No really, I know it's mine because https://git.kernel.org/pub/scm/linux/kernel/git/mpe/linux-fullhistory.git/tag/?h=v2.4.0 mentions me. Just clone that then change the upstream to the normal kernel.org one and git pull to current, makes "git annotate" WAY more useful.) And I've got portability.h to defend against libc breakage.

But I waited an extra 6 months to use "wait -n", AND I'm implementing a shell intended to have "wait -n" in it, and that not being _enough_ is... vexing.

enh-google:
> But that seems unrelated to running a newer build environment on that hardware?

that was what i was trying to say --- forcing people to buy new hardware makes me feel like a bad guy, but forcing people to take a _free_ OS upgrade (especially to a newer "LTS" version) ... has never struck me as particularly unreasonable.

given that CentOS 8 is just a few months away from its final "full" update (aiui from https://en.wikipedia.org/wiki/CentOS), it doesn't seem unreasonable to upgrade off the _previous_ version? (and that's ignoring that aiui CentOS is abandonware now anyway, so it's probably not just time to get a newer CentOS, but time to get a different OS entirely!)

PR #293, opened 2021-09-11
Minor readability changes to time.

prettykool:
Specifically, a newline was added, and tabs were used instead of spaces.
If you're curious, here's the original output of `time`, with the result of the sample input  `echo "Hello."` included:

```sh
Hello.
real 0.034699
user 0.000000
sys 0.000000
```
And here's the current output:
```sh
Hello.

real    0.030935
user    0.000000
sys     0.015625
```
That's it.


prettykool:
Patch added to mainline, as [5d9432ba05f12301939ad669ecf318a707fdcea8](https://github.com/landley/toybox/commit/5d9432ba05f12301939ad669ecf318a707fdcea8). Didn't realize this was still open until just now...

PR #297, opened 2021-10-26
tftpd: fix the parsing of options

nzmichaelh:
TFTP supports passing options such as 'blksize' as part of the initial
RRQ or WRQ packet. These options consist of a null-terminated name
followed by a null-terminated value. The current code segfaults due to
mis-calculating the address of the value.

Add a `next_token` helper that advances past the next null and also
detects advancing past the end of the packet.

Tested by running:

    atftp -g -r /toybox -l toybox --option "blksize 1500" --option "foo bar" localhost 8069

Signed-off-by: Michael Hope <mlhx@google.com>

### DIFF
--- a/toys/pending/tftpd.c
+++ b/toys/pending/tftpd.c
@@ -84,6 +84,21 @@ static void send_errpkt(struct sockaddr *dstaddr,
     perror_exit("sendto failed");
 }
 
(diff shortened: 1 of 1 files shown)


enh-google:
(merged directly by rob the other day.)

PR #323, opened 2022-03-04
Update lsusb.c

changyeonjo:
Because scan_uevent() returns a number of patterns it successfully filled, its result should be evaluated as successful when its return value equals to its len parameter.


landley:
The second change breaks lspci. (That one's supposed to be != to skip entries.)

I had a couple other pending changes so I rolled the first fix into 6f24967142db.

Thanks,

Rob

changyeonjo:
Thank you, landley.

enh-google:
thanks! pulled into AOSP with https://android-review.googlesource.com/c/platform/external/toybox/+/2009395

PR #326, opened 2022-03-16
Enable copy_file_range when it is available.

LongPingWEI:
Changes:
1. fix the mistake in scripts/genconfig.sh;
2. `syscall(__NR_copy_file_range ...)` is more suitable, as there is not a `copy_file_range` in bionic.
3. Even compiled with glibc, toybox cannot use [copy_file_range](https://man7.org/linux/man-pages/man2/copy_file_range.2.html) easily, as we must define _GNU_SOURCE before include unistd.h first time.

Test:
1. dd if=/dev/urandom of=src.bin bs=4M count=1000
2. ln -s toyobox cp
5. ./cp src.bin dst.bin
6. sha256sum src.bin dst.bin

### DIFF
--- a/lib/portability.c
+++ b/lib/portability.c
@@ -629,20 +629,25 @@ int get_block_device_size(int fd, unsigned long long* size)
 long long sendfile_len(int in, int out, long long bytes, long long *consumed)
 {
   long long total = 0, len, ww;
+  int copy_file_range = CFG_TOYBOX_COPYFILERANGE;
 
   if (consumed) *consumed = 0;
   if (in<0) return 0;
   while (bytes != total) {
     ww = 0;
     len = bytes-total;
-    if (bytes<0 || len>sizeof(libbuf)) len = sizeof(libbuf);
 
     errno = 0;
-#if CFG_TOYBOX_COPYFILERANGE
-    len = copy_file_range(in, 0, out, 0, bytes, 0);
-#else
-    ww = len = read(in, libbuf, len);
-#endif
+    if (copy_file_range) {
+      if (bytes<0) len = INT_MAX;
+      len = syscall(__NR_copy_file_range, in, 0, out, 0, len, 0);
+      if (len < 0 && errno == EINVAL)
+        copy_file_range = 0;
+    }
+    if (!copy_file_range) {
+      if (bytes<0 || len>sizeof(libbuf)) len = sizeof(libbuf);
+      ww = len = read(in, libbuf, len);
+    }
     if (len<1 && errno==EAGAIN) continue;
     if (len<1) break;
     if (consumed) *consumed += len;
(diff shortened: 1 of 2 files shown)


landley:
Did you run "make tests"? Because the cp tests are failing after this.

LongPingWEI:
> Did you run "make tests"? Because the cp tests are failing after this.
@landley I have fixed the problems caused by the first patch. "make tests" pass on my environment now.

There are two mistakes in the first patch:
1. I ignored that `sendfile_len` supports parameter `bytes` be negative;
2. `copy_file_range` cannot work when either `in` or `out` is not a regular file;

enh-google:
> * `syscall(__NR_copy_file_range ...)` is more suitable, as there is not a `copy_file_range` in bionic.

feel free to report bionic bugs like this at https://github.com/android/ndk --- i could have sworn we'd added this already, but you're right; all we have is a "seccomp hole" so that it's possible to make the syscall without being killed. i don't even appear to have filed a bug myself to add this...

landley:
I already applied it locally and I kept the syscall() because glibc requires you to #define GNU_GNU_ALL_HAIL_STALLMAN_FTAGHN to get the linux system call wrapper, which is silly. (I'd have poked you otherwise, but glibc was already broken. I made puppy eyes at Rich to add _ALL_SOURCE which just enables all the darn feature test macros, but glibc still wants to gate Linux system calls on The Hurd for some reason. Well, ok, we know the reason, it was https://sourceware.org/legacy-ml/libc-announce/2001/msg00000.html#:~:text=Stallman but it's still really stupid.)

But I haven't pushed the yet because https://github.com/torvalds/linux/commit/f16acc9d9b376 from 2019 talks about adding "basic sanity checks", which seems to boil down to capping the length at INT_MAX-PAGE_SIZE. That's well within the 7 year horizon and since the code's already in the read/write loop handling length returns, I thought I'd do a follow-up commit capping len at 1<<30 or so per copy_file_range() call?

landley:
Try 9b368059deec.

LongPingWEI:
> I already applied it locally and I kept the syscall() because glibc requires you to #define GNU_GNU_ALL_HAIL_STALLMAN_FTAGHN to get the linux system call wrapper, which is silly. (I'd have poked you otherwise, but glibc was already broken. I made puppy eyes at Rich to add _ALL_SOURCE which just enables all the darn feature test macros, but glibc still wants to gate Linux system calls on The Hurd for some reason. Well, ok, we know the reason, it was https://sourceware.org/legacy-ml/libc-announce/2001/msg00000.html#:~:text=Stallman but it's still really stupid.)
> 
> But I haven't pushed the yet because [torvalds/linux@f16acc9d9b376](https://github.com/torvalds/linux/commit/f16acc9d9b376) from 2019 talks about adding "basic sanity checks", which seems to boil down to capping the length at INT_MAX-PAGE_SIZE. That's well within the 7 year horizon and since the code's already in the read/write loop handling length returns, I thought I'd do a follow-up commit capping len at 1<<30 or so per copy_file_range() call?

Linux kernel will limit the max size to 0x7FFFF000 in `vfs_copy_file_range`.

Issue #328, opened 2022-03-23
make tests fails after 4fca350fb34c31e681fd6750692d6089b9810c3d


LongPingWEI:
PASS: tail noseek -c+ out of bonds
PASS: tail -c 12345 -n 3 bigfile
PASS: tail -n 3 -c 12345 bigfile
PASS: tail -f one
PASS: tail -f one two three
PASS: tail -F
PASS: tar create file
PASS: tar pass file
PASS: tar pass user
tar: bad group 'nobody'
tar: Not tar
FAIL: tar pass group
echo -ne '' | tar c --owner root --group nobody --mtime @0 file | LST
--- expected    2022-03-23 11:04:31.921672300 +0000
+++ actual      2022-03-23 11:04:31.931672300 +0000
@@ -1 +0,0 @@
--rw-rw-r-- root/nobody 0 1970-01-01 00:00 file
make: *** [Makefile:77: tests] Error 1

LongPingWEI:
@landley My environment Ubuntu 20.04 WSL.

LongPingWEI:
I found that this commit modified the perm of tar.test from 0644 to 0755.
If I set perm 0644 to tar.test, make tests can pass.

LongPingWEI:
The other issue in tar.test:
```
yes | (dd bs=$((1<<16)) count=1 status=none; dd bs=8192 seek=14 count=1 status=none; dd bs=4096 seek=64 count=5 status=none) > fweep
testing "sparse without overflow" "$TAR --sparse fweep | SUM 3" \
  "e1560110293247934493626d564c8f03c357cec5\n" "" ""

rm fweep
```
The length of `fweep`  differs from what generated by the same command on Ubuntu.
```
weilongping@Desktop-of-LongPing:~/toybox$ ls -lh ./generated/testdir/testdir/fweep
-rw-rw-r-- 1 weilongping weilongping 381K Mar 23 20:38 ./generated/testdir/testdir/fweep
weilongping@Desktop-of-LongPing:~/toybox$ ls -lh ./fweep
-rw-r--r-- 1 weilongping weilongping 428K Mar 23 20:33 ./fweep
```

enh-google:
> ```
> id nobody
> uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)
> ```
> 
> but tar.test try to use nobody as a group,
> 
> ```
> skipnot id nobody >/dev/null
[...]

PR #329, opened 2022-03-23
Fix failures in tar.test

LongPingWEI:


LongPingWEI:
Fix the two problems metioned by [make tests fails after 4fca350fb34c31e681fd6750692d6089b9810c3d](https://github.com/landley/toybox/issues/328)

PR #336, opened 2022-04-27
brctl: Fix addif and defif

ek9852:
In br_addif and br_delif, the bridge
name is not passed to the ioctl.


landley:
Commit 80206bb53765

PR #339, opened 2022-05-03
Check return value of fscanf.

jfgoog:
Eliminates an unused result warning when compiling with clang.

### DIFF
--- a/toys/pending/sh.c
+++ b/toys/pending/sh.c
@@ -709,14 +709,18 @@ static struct sh_vars *setvar_long(char *s, int freeable, struct sh_fcall *ff)
   if (ss[*ss=='+']!='=') {
     error_msg("bad setvar %s\n", s);
     if (freeable) free(s);
+
     return 0;
   }
 
   // Add if necessary, set value, and remove again if we added but set failed
   if (!(was = vv = findvar(s, &ff))) (vv = addvar(s, ff))->flags = VAR_NOFREE;
-  if (!(vv = setvar_found(s, freeable, vv))) {
+  if (!setvar_found(s, freeable, vv)) {
     if (!was) memmove(vv, vv+1, sizeof(ff->vars)*(--ff->varslen-(vv-ff->vars)));
-  } else cache_ifs(vv->str, ff);
+
+    return 0;
+  }
+  cache_ifs(vv->str, ff);
 
   return vv;
 }


landley:
I'm not seeing this warning with android-ndk-r21d. Which version can I reproduce this with?

The logic here zeroes len and ll before each fscanf(), so if either isn't read they stay zero. The source of the data is a host toysh instance, and verifying that is the reason for the pid!=getpid() || ppid!=getppid() check above. If the data in the pipe is corrupted, error_exit() is probably the appropriate response, which is what happens for a short read() later in the loop: if we don't get as many bytes of variable data as we expected, we abort the subshell, so that's where a short read gets checked.

landley:
The previous write to that fd was write() with perror_msg() but not perror_exit(), because failure to set selinux context is "error message and eventual nonzero exit code" but not "abort archive extract immediately".

It's possible the zero length write to reset context can never return an error so xwrite() would merely add NOP checking, but I don't know what the reset returns if the previous write had an error? If it repeats the error, xwrite() would abort the archive processing.

landley:
And it's not trying to "fool" gcc, the typecast is saying that we explicitly do not care about this return value. It wasn't overlooked, we explicitly made the decision _not_ to check it because there's nothing useful to do with the information.

enh-google:
yeah, but you know my attitude is "this can't fail ... so if this _does_ fail that would be _massively_ surprising, so xwrite() seems appropriate". i'd much rather have a loud failure showing up in the crash collection infrastructure than just weird side-effects that no-one notices. (even/especially if it's a kernel bug --- there are few things harder to track down, and libc and toybox are exactly the places where we end up finding those bugs :-( )

jfgoog:
Confirmed that `make CFLAGS='-Wunused-result -D_FORTIFY_SOURCE=2'` triggers the warning:

```
toys/net/netstat.c:103:3: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
  fgets(toybuf, sizeof(toybuf), fp);
  ^~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~
.toys/net/netstat.c:173:3: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
  fgets(toybuf, sizeof(toybuf), fp);
  ^~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~
toys/net/netstat.c:254:3: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
  fgets(toybuf, sizeof(toybuf), fp);
  ^~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~
3 warnings generated.
toys/other/watchdog.c:34:3: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
  write(TT.fd, "V", 1);
  ^~~~~ ~~~~~~~~~~~~~
toys/other/watchdog.c:48:5: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
    write(TT.fd, "", 1);
    ^~~~~ ~~~~~~~~~~~~
2 warnings generated.
toys/pending/sh.c:3841:5: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
    fscanf(fp, "%u %lu%*[^\n]", &len, &ll);
    ^~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

(My default compiler is clang, but `make CC=gcc CFLAGS='-Wunused-result -D_FORTIFY_SOURCE=2'` produces similar warnings.)

PR #342, opened 2022-05-07
Diff line format options.

jfgoog:
--unchanged-line-format
--old-line-format
--new-line-format

We want to use Toybox to make AOSP builds closer to hermetic, and these
flags are used here:
https://android.googlesource.com/platform/build/+/refs/tags/android-wear-9.0.0_r20/target/product/vndk/Android.mk


landley:
Commit 9f5fb9ae9af1

PR #344, opened 2022-05-17
Handle diffing FIFOs.

jfgoog:
Also exit immediately if a seek fails.

In addition to the supplied test, also tested manually with:

./toybox diff <(echo -e "1\n2") <(echo -e "1\n3")

However, the "testing" function in runtest.sh doesn't handle such
syntax.

### DIFF
--- a/tests/diff.test
+++ b/tests/diff.test
@@ -42,3 +42,16 @@ testing "--strip-trailing-cr on" "diff -u --strip-trailing-cr a b" "" "" ""
 echo -e "1\n2" > aa
 echo -e "1\n3" > bb
 testing "line format" "diff --unchanged-line-format=U%l --old-line-format=D%l --new-line-format=A%l aa bb" "U1D2A3" "" ""
+
+mkfifo fifo1
+mkfifo fifo2
+echo -e "1\n2" > fifo1&
+echo -e "1\n3" > fifo2&
+expected='--- fifo1
++++ fifo2
+@@ -1,2 +1,2 @@
+ 1
+-2
++3
+'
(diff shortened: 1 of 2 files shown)


enh-google:
> Also exit immediately if a seek fails.
> 
> In addition to the supplied test, also tested manually with:
> 
> ./toybox diff <(echo -e "1\n2") <(echo -e "1\n3")
> 
> However, the "testing" function in runtest.sh doesn't handle such syntax.

(which is probably a good thing for us, because iirc mksh doesn't support it anyway. iirc i had to rewrite some toybox tests to _not_ use that.)

jfgoog:
Fixed diff so that line formats can be empty.

landley:
I use diff <(thingy) <(thingy) all the time, diff NOT supporting that was one of the todo items for the command keeping it in pending...

And I'm actively working on getting toysh to run the test suite. Trying to close a couple other tabs first, but it's "next release or maybe the one after" territory...

landley:
Applied.

landley:
When a --flag is seen the bit in toys.optflags gets set, ala { if (FLAG(foo_line_format)) blah(); }, and also TT.foo_line_format would point to a "" string instead of being NULL when the =argument is supplied.

You can use ; in the optstring for the argument to be optional (in which case it can only be supplied as a --longopt=arg rather than --longopt arg).

jfgoog:
Yes, I added `;` to the optstring in my second commit to this PR, which you merged.

I didn't check `if (FLAG(foo_line_format))` though, because I think it's enough to check if the char ptr is null, and we're already doing that.

PR #347, opened 2022-05-23
Use extglob rather than regex match operator.

jfgoog:
Not all shells support =~. Android, for example, uses mksh, which does
not.

Change is from suggestion at http://www.mirbsd.org/mksh-faq.htm#regex-comparison


landley:
They're not the same thing, would break sh.test "sh functions need block" test for example (looking for [Ss]yntax).

I replaced it with a call to grep -q in commit 1c7412199a7e

PR #348, opened 2022-05-24
Handle test.test being run as root.

jfgoog:
Root always has read and write access.

Tested with `sudo make test_test`


landley:
Commit 9febe910b6c1

PR #349, opened 2022-05-24
Use $(id -u) instead of $UID to detect root.

jfgoog:
We can't depend on a value being set for $UID.


landley:
Commit fae6c10e6b0a

PR #350, opened 2022-06-01
Ignore text editor generated files.

mahiuchun:


landley:
I like to see those so I can delete them. (And my editor backups are .filename.sw?)

mahiuchun:
For me it was emacs with `*~` (by default, it remains after the editor is exited normally), then I saw such a section in linux's `.gitignore`. `*.orig` is actually for patch i guess.

If you don't like this then I could close it. I haven't seen any previous attempt for something like this.

Issue #362, opened 2022-08-13
linux filesystem conflict with glibc >=2.36 


wael444:
to recreate:
```
git clone https://github.com/landley/toybox
cd toybox
make defconfig
make
```
output:
```make
scripts/make.sh
Library probe
generated/{Config.in,newtoys.h,flags.h,globals.h,tags.h,help.h}
Compile toybox.................In file included from /usr/include/stdio.h:906,
                 from ./toys.h:30,
                 from lib/password.c:8:
In function 'fprintf',
    inlined from 'update_password' at lib/password.c:148:21:
/usr/include/bits/stdio2.h:79:10: warning: pointer 'nfp' may be used after 'fclose' [-Wuse-after-free]
   79 |   return __fprintf_chk (__stream, __USE_FORTIFY_LEVEL - 1, __fmt,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   80 |                         __va_arg_pack ());
      |                         ~~~~~~~~~~~~~~~~~
lib/password.c: In function 'update_password':
lib/password.c:144:3: note: call to 'fclose' here
  144 |   fclose(nfp);  // automatically unlocks
      |   ^~~~~~~~~~~
At top level:
cc1: note: unrecognized command-line option '-Wno-string-plus-int' may have been intended to silence earlier diagnostics
....................toys/net/netstat.c: In function 'display_routes':
toys/net/netstat.c:254:3: warning: ignoring return value of 'fgets' declared with attribute 'warn_unused_result' [-Wunused-result]
  254 |   fgets(toybuf, sizeof(toybuf), fp);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
toys/net/netstat.c: In function 'show_ip':
toys/net/netstat.c:103:3: warning: ignoring return value of 'fgets' declared with attribute 'warn_unused_result' [-Wunused-result]
  103 |   fgets(toybuf, sizeof(toybuf), fp);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
toys/net/netstat.c: In function 'show_unix_sockets':
toys/net/netstat.c:173:3: warning: ignoring return value of 'fgets' declared with attribute 'warn_unused_result' [-Wunused-result]
  173 |   fgets(toybuf, sizeof(toybuf), fp);
[...]

PR #369, opened 2022-08-21
fix glibc 2.36 build

wael444:
solves #362 
note: i don't know C or what i am doing, i just want this to be fixed.


kraj:
You want to upgrade glibc to latest on 2.36 release branch. This patch is not required

wael444:
oh my god thank you

PR #373, opened 2022-09-08
portability: Fix timer_settime_wrap for 32bit systems with 64bit time_t

kraj:
glibc does not define SYS_timer_settime if the 32bit syscall is not available, new architectures like riscv32 has defaulted to 64bit time_t from get go and avoided wiring 32bit syscall, therefore alias it to 64bit version here


landley:
Applied as 9c9d3fa7ead5

PR #374, opened 2022-09-14
fmt: Define GLOBALS 'width' as long, not int

frikiluser:
This change fixes fmt toy for Linux/s390x.
Without that change width value is always 0 on that arch.

PS:
'make tests' pass after that change on Debian/s390 for defconfig.
All other Debian archs pass tests even without this fix.


landley:
Commit d4ed83762519

PR #391, opened 2022-10-20
Fix undefined behaviour in sed.

fmayer:
This UB breaks builds of sed with HWASan enabled in Android, because the check of the `while (command)` is optimized away.

Adding to NULL pointer and subtracting back is undefined behaviour, as it is pointer arithmetic beyond the bounds of the object. The compiler is free to assume that `ptr - 1` is non-NULL.

### DIFF
--- a/toys/posix/sed.c
+++ b/toys/posix/sed.c
@@ -156,6 +156,9 @@ struct sedcmd {
   char c; // action
 };
 
+// Placeholder sedcmd for the end of commands.
+static struct sedcmd nullcmd;
+
 #define SFLAG_i 1
 #define SFLAG_g 2
 #define SFLAG_p 4
@@ -267,9 +270,9 @@ static void sed_line(char **pline, long plen)
   }
   TT.count++;
 
-  // The restart-1 is because we added one to make sure it wasn't NULL,
-  // otherwise N as last command would restart script
-  command = TT.restart ? ((struct sedcmd *)TT.restart)-1 : (void *)TT.pattern;
+  command = TT.restart ? ((struct sedcmd *)TT.restart) : (void *)TT.pattern;
+  if (command == &nullcmd)
+    command = NULL;
   TT.restart = 0;
 
   while (command) {
@@ -427,14 +430,14 @@ static void sed_line(char **pline, long plen)
       toybuf[off++] = '$';
       emit(toybuf, off, 1);
     } else if (c=='n') {
-      TT.restart = command->next+1;
+      TT.restart = command->next ? command->next : &nullcmd;
 
       break;
     } else if (c=='N') {
       // Can't just grab next line because we could have multiple N and
       // we need to actually read ahead to get N;$p EOF detection right.
       if (pline) {
-        TT.restart = command->next+1;
+        TT.restart = command->next ? command->next : &nullcmd;
         extend_string(&line, TT.nextline, len, -TT.nextlen);
         free(TT.nextline);
         TT.nextline = line;


landley:
That would be a broken compiler.

landley:
Is your compiler breaking on the -1 or on the +1? Because my first glance was to do:

+      // Work around insane broken "optimizer" refusing to reliably inc ptr
+      TT.restart = (void *)(((unsigned long)command->next)+1);

But you're saying it's the -1 that needs the (unsigned long) to get it to shut up? (I haven't reproduced this failure, my compiler isn't broken...)

landley:
Or does while ((unsigned long)command) work as a more localized fix? It sounds like it's getting the math right (it would kind of have to) and it's the dead code elimination that needs a newspaper to the snout...

landley:
If all else fails I can make command volatile...

enh-google:
(courtesy of someone else on the team), it's that the optimizer knows (because the standard tells it so) that "p-1" can never turn a non-nullptr into nullptr.

Take, for example:
```
int glob;
void g(int *p);
void f(int *p) {
  int *q = p ? p - 1 : &glob;
  if (q) g(q);
}
```
At -O2, the function call becomes unconditional:
```
        adrp    x8, glob
        add     x8, x8, :lo12:glob
        sub     x9, x0, #4
        cmp     x0, #0
        csel    x0, x8, x9, eq
        b       _Z1gPi
```

eugenis:
oh it's even simpler:
```
void g();
void f(int *p) {
  if (!(p - 1))
    g();
}
```
compiles to
```
        ret
```
Because a pointer must always point either to a live object, first byte past one, or nullptr. And there is no case where (p-1) is nullptr.

fmayer:
6.5.6 https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf

> When an expression that has integer type is added to or subtracted from a pointer, the
> result has the type of the pointer operand. If the pointer operand points to an element of
> an array object, and the array is large enough, the result points to an element offset from
> the original element such that the difference of the subscripts of the resulting and original
> array elements equals the integer expression. In other words, if the expression P points to
> the i-th element of an array object, the expressions (P)+N (equivalently, N+(P)) and
> (P)-N (where N has the value n) point to, respectively, the i+n-th and i−n-th elements of
> the array object, provided they exist. Moreover, if the expression P points to the last
> element of an array object, the expression (P)+1 points one past the last element of the
> array object, and if the expression Q points one past the last element of an array object,
> the expression (Q)-1 points to the last element of the array object. If both the pointer
> operand and the result point to elements of the same array object, or one past the last
> element of the array object, the evaluation shall not produce an overflow; otherwise, the
> behavior is undefined.

landley:
enh: which standard? (Is this something else that moved going from c99 to c11? I read c90 cover to cover back in the day, and had many years to get familiar with c99, but have not sat down and spent a week staring at c11 looking for booby traps.) You're saying C11 no longer has pointers, it only has references? Because pointers can point to anything: that's what makes them pointers. It's dereferencing them that's fraught. I can (void *)0x123456789abcdef and the result is a pointer.

Neither my distro's gcc nor the musl-cross-make toolchains exhibited this, so I downloaded NDK 25b and used the aarch64 cross compiler from there to reproduce, and yes I'm seeing the same failure. I stuck "volatile" on both of the local and the function argument and the compiler still miscompiled them:  this compiler's "volatile" isn't.

Typecasting to unsigned long and back worked fine though, adding:
14: 40 00 00 b4  	cbz	x0, 0x1c <f+0x1c>

What test did you use to see different behavior from sed for this? Did "make test_sed" fail? If so, how did nobody notice until the updated toolchain had already shipped in the NDK?... (Or was it one of the sed tests I added recently?)

I use (void *)1 as a placeholder I can explicitly test for in several places in toybox (because it's not null but is in the zero page so will never match any pointer returned by malloc/mmap/etc). That diff algorithm I checked into a branch increments pointers in an array to tag them, because I know pointers returned by malloc() have to be word aligned so the bottom bit is always otherwise zero (and then the consumer of the tagged pointers decements them so they can be freed again afterwards). This is very much NOT the only place compiler developers expanding their confusion of C's pointers with C++'s references is going to break code that's been working for years.

If I have to go back to c99 and rip out the c11 features and --std=nonbroken to get behavior that doesn't break regularly, I can do that? I'd really rather have a -fstop-doing-that compiler flag ala -fno-strict-aliasing and -funsigned-char...

Sigh: I pushed commit 3d097856ed3b to "branch2", is that a sufficient fix for this issue? (It passed make test_sed with misc bin qemu running the binaries, but I'm not sure what that proves...)

landley:
Deleted the branch and commited it to master. If it's not a sufficient fix I can check something else in on top of it.

fmayer:
I can confirm that 3d097856ed3b15a29c17acf9aa7139698357a486 fixed the crash I was seeing.

PR #415, opened 2023-03-20
rename itoa to toybox_itoa in seq.c, to avoid clashes

justdan96:
As title, this is to avoid clashing with systems where `itoa` is already defined. If the name is not satisfactory then just let me know of a more appropriate one.

### DIFF
--- a/toys/lsb/seq.c
+++ b/toys/lsb/seq.c
@@ -54,7 +54,7 @@ static double parsef(char *s)
 
 // fast integer conversion to decimal string
 // TODO move to lib?
-static char *itoa(char *s, int i)
+static char *toybox_itoa(char *s, int i)
 {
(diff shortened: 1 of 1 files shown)


landley:
For the record, which build environment breaks?

justdan96:
I'm using Cygwin on Fedora - I know it's not supported by toybox but I figured other build environments may encounter the same issue.

landley:
I can move itoa() to portability.c, is there a cygwin #ifdef I can check for?

justdan96:
Yes, you can use `#if defined(__CYGWIN__)` to just check for Cygwin and the `itoa` functions are hidden behind `#if __MISC_VISIBLE`, which is enabled by default in Cygwin.

PR #416, opened 2023-03-20
change xrunread parameter name to to_stdin, matching xwrap.c

justdan96:
As title, it looks like the first part of this was done in https://github.com/landley/toybox/commit/3dae8ebfe590a1b3ae86d2dc9d3e206357fd2eba, this is just to complete the change.


landley:
Commit 46e64fb415d9

PR #441, opened 2023-07-03
Fix GLOBALS() for nsenter

jooyunghan:
When -C was added in d0c52934a39a5795af12bbf904a093306e38981e, the number for the arguments were not modified.

-t argument is not set correctly.

Change-Id: I17f448c5d29b1cafb1a1e5226025eaa96cdbb92d


landley:
Commit a6ad3659424b

PR #463, opened 2023-10-27
ip: improve 'ip link set' parsing

man-gc:
Some utilities expect that the order of parameters does not matter since it is the behavior of iproute2 utility. So, fix parsing to provide the same behavior for toybox.


man-gc:
Closing, because patch is already merged

PR #481, opened 2024-02-18
Fix stty not setting baud rate to 57600

sukem:
When attempting to set the baud rate to 57600 using stty, the command is ignored. This is because baud rates of 57600 and higher require a special flag (CBAUDEX) to indicate extended baud rates.


landley:
Commit 5c8cd70ae036

PR #482, opened 2024-02-19
Fix tests for getfattr/setfattr

tweksteen:
The test system may have extended attributes set by default (for instance, security.selinux on Android). Support these by only testing part of the output.

This is similar to tests/chattr.test which filters the output.

### DIFF
--- a/tests/getfattr.test
+++ b/tests/getfattr.test
@@ -4,14 +4,23 @@
 
 #testing "name" "command" "result" "infile" "stdin"
 
+function clean()
+{
+  # The filesystem may include some extended attributes by default (for
+  # instance, security.selinux). Skip them.
+  grep -v "security\."
+}
+
 mkdir attrs
 touch attrs/file
 setfattr -n user.empty attrs/file
 setfattr -n user.data -v hello attrs/file
+setfattr -n user.more -v world attrs/file
 
-# The filesystem may include some extended attributes by default (for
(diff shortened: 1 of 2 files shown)


tweksteen:
@enh-google fyi

tweksteen:
That's right, the kernel may add some of these labels depending on the system configuration (that's why just filtering out `security.selinux` might not be robust enough).

I'm not sure how much coverage you preferred, but yes the `-n` option is more precise. I've updated the commit to use it in all test cases. Thanks.

enh-google:
that seems like a bad idea? we already had tests for `-n`, but now we don't have tests for the non-`-n` path? (personally i'd have just filtered out the security cruft. realistically i don't think we're going to see others; smack just didn't get the same traction afaict, and even if it did, that's just one more thing to filter out. filtering also matches what we've done in similar cases in other tests...)

tweksteen:
There is `security.ima` that could realistically be in use? From reading xattr(7) man page, I think other namespaces are unlikely to be populated. Maybe we could filter out anything in the `security` namespace (that is, matching `security.`)?

enh-google:
yes, that's even better if others are already using the `security.` namespace!

tweksteen:
I didn't realise the previous version of the commit got merged. Just in case, there is an updated version following the conversation above. Thanks.

landley:
First time I've applied the same pull request twice. Commit f79b72761f49

PR #484, opened 2024-03-08
Adjust documentation to real possible values

gizmomogwai:


landley:
Should the -n option check the value in the option string? What _is_ the actual range...

According to https://github.com/torvalds/linux/blob/master/include/linux/kern_levels.h it's 0-7, for -n we're calling logctl(8, 0, TT.n), musl's klogctl.c is a syscall wrapper that doesn't adjust the values, linux/kernel/printk/printk.c has the syscall entry point call do_syslog(type, buf, len, SYSLOG_FROM_READER); and that function earlier in the same C file... is using macros, grep to find... "include/linux/syslog.h:#define SYSLOG_ACTION_CONSOLE_LEVEL  8" and in the case statement that does if (len < 1 || len > 8) return -EINVAL;

So yes, 1-8.

*shrug* Applied as-is, the kernel catches it if we don't check. Commit 38505e0ab9a1

gizmomogwai:
Actually I was fishing, if those values do make sense at all. My embedded device's linux kernel accepts values of 1-8 (inclusive). Might be explained as show all messages of level lower than what is given as argument. Also what e.g. https://man7.org/linux/man-pages/man1/dmesg.1.html makes clear, is that its only the level of messages going to console that are influenced by -n. Also other implementations e.g. in ubuntu prohibit the use of `-r` and `-n` at the same time.

For toybox I think its great as it is now though.

PR #485, opened 2024-03-14
hexdump: fix NUL character handling

man-gc:
When dumping a file containing some NUL characters, the display is wrong because all remaining characters on the line are replaced by NUL. This error comes from the fact that the copy of the read buffer into the line buffer is done using strncpy() which handles NUL character as end of string and pads the destination buffer with NUL afterwards.

To avoid data manipulation during buffer copy, simply replace strncpy() with its memory equivalent memcpy() that does not interpret the read data.

### DIFF
--- a/toys/pending/hexdump.c
+++ b/toys/pending/hexdump.c
@@ -102,7 +102,7 @@ void do_hexdump(int fd, char *name)
     continue;
 
 newline:
-    strncpy(TT.linebuf+(TT.bc%16), toybuf, TT.len);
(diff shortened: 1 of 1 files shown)


landley:
Applied (commit cab0b6653827).

The reason this command is still in pending (hasn't been cleaned up and promoted) is we already have THREE hex dumpers (od, xxd, and hexedit), and this is a fourth that doesn't share any code with the other three. I haven't come up with an obvious way to make them share code usefully, but am annoyed at having 4 implementations of the same functionality in the tree. (I should just swallow the redundancy, but... Ouch.)

PR #505, opened 2024-06-11
portability for {Free,Net,Open,}BSD

agreppin:


landley:
Touches more files at once than I'm happy with. (The test suite plumbing tends to use SKIP rather than if around the tests, to avoid running different NUMBERS of tests on different targets, and silently skipping tests without saying "SKIP:")...

Came to the web interface to see if the 1/2 that got squished together by wget 505.patch provided better granularity, but doesn't look like it. (I can break it up easily enough, but then either I'm faking metadata or don't attribute it properly...)

I'll take a swing at it...

agreppin:
please note that I only tested with this config:
```
CONFIG_TOYBOX_FORK=y
CONFIG_PRINTF=y
CONFIG_SORT_FLOAT=y
CONFIG_BASE64=y
CONFIG_HELP=y
CONFIG_GZIP=y
CONFIG_GUNZIP=y
CONFIG_ZCAT=y
CONFIG_TOYBOX=y
CONFIG_TOYBOX_LSM_NONE=y
CONFIG_TOYBOX_FLOAT=y
CONFIG_TOYBOX_HELP=y
CONFIG_TOYBOX_ZHELP=y
CONFIG_TOYBOX_DEBUG=y
CONFIG_TOYBOX_UID_SYS=100
CONFIG_TOYBOX_UID_USR=500
```

lib/portability.h
reactive-firewall, line 1: :+1: LGTM; Regarding `lib/portability.h` with the caveat of two minor style suggestions. See inline suggestions for details.

scripts/portability.sh
```diff
@@ -28,6 +37,11 @@ fi
 # And ones only gcc produces
   CFLAGS+=" -Wno-restrict -Wno-format-overflow"
 
+case "$(uname)" in
+'NetBSD'|'OpenBSD')
+  unset ASAN;;
```
reactive-firewall, line 42: :confused: does OpenBSD really have issues with address sanitation? Please explain for those of us who are uninitiated. (:pray: with small, gentle words please)

scripts/portability.sh
```diff
@@ -8,6 +8,10 @@ then
   exit 1
 fi
 
+case ${OD-} in
+'') OD=$(command -v god || command -v ggod || echo od);; # use od from GNU coretutils
```
reactive-firewall, line 12: :no_good: This change concerns me by deliberately introducing a reliance on "GNU coretutils" and places the original `od` command at the end of the chained-or-logic lending to a preference on GNU.

Perhaps at least consider maintaining `od` at the start of the logic. Otherwise please add a comment for _why_ this logic is needed for those of us yet uninitiated.

tests/zcat.test
```diff
@@ -4,6 +4,8 @@
 
 #testing "name" "command" "result" "infile" "stdin"
 
+HEAD=$(command -v ghead || echo head) # use head from GNU coreutils
```
reactive-firewall, line 7: :confused: Similar to before in `scripts/portability.sh` this preference for GNU variants concerns me. Please consider maintaining the preference for the original variant in the chained-or-logic. Otherwise please enlighten us uninitiated about why NetBSD would require this.

main.c
```diff
@@ -151,6 +151,10 @@ void check_help(char **arg)
   }
 }
 
+#ifdef __NetBSD__
```
reactive-firewall, line 154: :broom: NIT - Consider a consistent macro style

Similar to other files; TL;DR

Please consider:

```suggestion
#if defined(__NetBSD__) && !defined(uselocale)
// use a No-OP for NetBSD instead of uselocale() for compatability 
```

main.c
reactive-firewall, line 1: :shrug: LGTM; Regarding main.c the no-op implementation for `uselocale` is straight forward. See suggestion on style inline.

reactive-firewall (commented):
I did a partial Review in hopes that it helps @landley, and the changes to `lib/protability.*` are trivial, I'll recommend further review of scripts specificly in the case of toybox's long-term plans.

See specific comments inline.

:bow: Hope this helps.

agreppin:
Thanks for the review, I will rework that PR, when time permits.

lib/portability.c
```diff
@@ -644,6 +644,10 @@ static int check_copy_file_range(void)
 #endif
 }
 
+#ifdef __OpenBSD__
+#define syscall(...) -1
+#endif
```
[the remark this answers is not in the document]
agreppin: syscall(2) is not available anymore on OpenBSD for common mortals
https://undeadly.org/cgi?action=article;sid=20231213062827

scripts/portability.sh
agreppin: not all `/bin/od` are equals, the test suite seems biased toward a specific implementation of `od`.

tests/zcat.test
agreppin: I don't remember now, seems a similar issue ... some tests fails with `head` and not with `ghead`.

scripts/portability.sh
reactive-firewall:
```suggestion
'') OD=$(command -v god || command -v ggod || echo od);; # use od from GNU coretutils for GNU biased tests
```

main.c
reactive-firewall: :+1: LGTM; Regarding `main.c` the no-op implementation for `uselocale` is straight forward, and targeted at NetBSD.

scripts/portability.sh
reactive-firewall: :bow: Thanks for the explanation, I was unaware.

:speak_no_evil: and I can almost hear the musl libc devs strictly claiming the lack of a check here is a bug too . :roll_eyes:

perhaps consider a comment:
```suggestion
  #TODO: add compiler check for '-fsanitize=address' (e.g., $CC -h 2>1 | grep 'args')
  unset ASAN;; # otherwise assume no compiler support by default
```

tests/zcat.test
reactive-firewall: :thinking: Odd; unless `ghead` expands the `2.gz` file before reporting the first 10 bytes (e.g., `$HEAD -c 10 2.gz`) I don't see why a BSD-Like `head` would behave different from `ghead` in this use-case.

perhaps a `-o pipefail` somewhere or similar side-efect issue :shrug:

---

:shrug: LGTM; Using a variable is maintainable enough.

reactive-firewall (commented):
:+1: LGTM; See suggested NIT suggestion/comments inline.

@landley this is probably ready for you to review.

:bow: I hope this helps.

PR #509, opened 2024-07-18
Fix ls -h total size bug

benny8453123:
With the -h option, the value should not be translated to block count before converting to a human-readable format.


benny8453123:
commit a new pull request in: https://github.com/landley/toybox/pull/510

PR #516, opened 2024-08-31
wget utility: invalid path traversal sanitization

LinRaymond2006:
Hello @landley , 
I wrote you an email about two weeks ago, but there's no response, so I open a pull request here.
### detail
If a Content-Disposition header is received and --output-document (or simply -O) is not specified, the creation of an arbitrary file can be triggered (not an overwrite of any existing files).
I think it's more like an undesirable behavior, rather than a vulnerability, since:
1. it requires user interaction and a request to an attacker-controlled server.
2. it does not overwrite any file, and without much information about the victim,  it's very unlikely to pose a serious threat to integrity of victim's machine.
3. executable bit of the output file is not set, there's no direct way for the user to execute the downloaded file.
4. I did see an attempt to validate the Content-Disposition variable (which seems not functioning). So instead of returning an error when a `/` is encountered, the patch ignores characters before last `/` (if any).

### reference
[rfc 2183 section 5](https://datatracker.ietf.org/doc/html/rfc2183#section-5)
[rfc 2616 section 15.5](https://datatracker.ietf.org/doc/html/rfc2616#section-15.5)
[rfc 2616 section 19.5.1](https://datatracker.ietf.org/doc/html/rfc2616#section-19.5.1)


LinRaymond2006:
Here's the proof-of-concept for your reference.
https://github.com/user-attachments/assets/0f102ec5-5fea-4c38-b8f7-79c7029bb2bd

PR #518, opened 2024-09-08
Use POSIX-compliant `od -A` argument value

andrewliebenow:
POSIX only specifies the characters 'd', 'o', 'x', and 'n' as valid `od -A` argument values. [1] Pedantic implementations of `od` may expect the argument value to be a single character instead of ignoring these extra characters. [2]

[1] https://pubs.opengroup.org/onlinepubs/9799919799/utilities/od.html
[2] https://github.com/uutils/coreutils/pull/6674


landley:
What od implementation did you use that broke?

andrewliebenow:
@landley The implementation in the uutils coreutils project. I linked the PR I opened to make it less picky.

landley:
I should do a FAQ entry for scripts/prereq/build.sh. Commit 3bbc31c78b41 explains it though. (Toybox can build its own prerequisites.)

andrewliebenow:
Oh, that's cool. Thanks.

PR #524, opened 2024-10-22
Add hwdata path to lsusb/lspci

firasuke:
Nowadays `/usr/share/hwdata` is considered a relatively common location to store the ID databases `pci.ids` and `usb.ids` and others.


landley:
Commit 2f07d87b26e8

PR #544, opened 2025-03-09
Remove unwanted perror_msg() when using tar --selinux

luk1337:
Doesn't seem to do anything useful and even seems to crash on my system.


landley:
On 3/9/25 11:57, luk1337 wrote:
> Doesn&#39;t seem to do anything useful and even seems to crash on my system.

Try commit 2cc5e25fb107

luk1337:
looks good now.

PR #545, opened 2025-03-09
Fix tar SELinux header

luk1337:
When comparing with files created by GNU tar, I noticed the following difference:

61 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\x00\n
->
60 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\n

After making toybox tar match GNU one, it no longer complains about bad header when trying to list the files inside.

Test: tar --selinux -cvf /tmp/a.tar /tmp/a
      tar --list -f /tmp/a.tar

### DIFF
--- a/toys/posix/tar.c
+++ b/toys/posix/tar.c
@@ -403,8 +403,8 @@ static int add_to_tar(struct dirtree *node)
       if (len>999999 || (sz && len>sz)) len = -1, errno = E2BIG;
       if (buf || len<1) {
         if (len>0) {
-          strcpy(buf+start+sz, "\n");
-          write_prefix_block(buf, start+sz+2, 'x');
+          strcpy(buf+start+sz-1, "\n");
+          write_prefix_block(buf, start+sz, 'x');
         } else if (errno==ENODATA || errno==ENOTSUP) len = 0;
         if (len) perror_msg("getfilecon %s", name);
 
@@ -415,7 +415,7 @@ static int add_to_tar(struct dirtree *node)
       // Allocate buffer. Length includes prefix: calculate twice (wrap 99->100)
       temp = snprintf(0, 0, "%d", sz = (start = 22)+len+1);
       start += temp + (temp != snprintf(0, 0, "%d", temp+sz));
-      buf = xmprintf("%u RHT.%s=%.*s", start+len+1, sec, sz = len, "");
+      buf = xmprintf("%u RHT.%s=%.*s", start+len, sec, sz = len, "");
     }
   }
 


landley:
On 3/9/25 12:04, luk1337 wrote:
> When comparing with files created by GNU tar, I noticed the following difference:
> 
> 61 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\x00\n -&gt;
> 60 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\n
> 
> After making toybox tar match GNU one, it no longer complains about bad header when trying to list the files inside.

Applied, but I'd like to add a test.

> Test: tar --selinux -cvf /tmp/tmp.sSP3udSmQs/a.tar /tmp/tmp.sSP3udSmQs/a
>        tar --list -f /tmp/tmp.sSP3udSmQs/a.tar

Could you send me an "a.tar"? (My devuan filesystem hasn't got xattrs on 
anything by default.)

I just need a small one file tarball I could add to the regression test 
suite, containing a file I can extract and tar up again where the 
resulting header block should be the same afterwards.

I've already got plumbing to override/regularize user and timestamp and 
so on in tests/tar.test already. I'm not quite sure how to teach the 
test plumbing to check whether the current filesystem has xattr support 
so it can skip the tests, but that's my problem not yours.

Thanks,

Rob

luk1337:
> On 3/9/25 12:04, luk1337 wrote: When comparing with files created by GNU tar, I noticed the following difference: 61 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\x00\n -&gt; 60 RHT.security.selinux=unconfined_u:object_r:user_tmp_t:s0\n After making toybox tar match GNU one, it no longer complains about bad header when trying to list the files inside.
> Applied, but I'd like to add a test.
> Test: tar --selinux -cvf /tmp/tmp.sSP3udSmQs/a.tar /tmp/tmp.sSP3udSmQs/a tar --list -f /tmp/tmp.sSP3udSmQs/a.tar
> Could you send me an "a.tar"? (My devuan filesystem hasn't got xattrs on anything by default.) I just need a small one file tarball I could add to the regression test suite, containing a file I can extract and tar up again where the resulting header block should be the same afterwards. I've already got plumbing to override/regularize user and timestamp and so on in tests/tar.test already. I'm not quite sure how to teach the test plumbing to check whether the current filesystem has xattr support so it can skip the tests, but that's my problem not yours. Thanks, Rob

[gnu.zip](https://github.com/user-attachments/files/19157390/gnu.zip)
[toybox.zip](https://github.com/user-attachments/files/19157391/toybox.zip)

(rename zip - >tar since GitHub doesn't allow .tar extension for whatever reason)

PR #546, opened 2025-04-11
Fix out of bounds error

fmayer:
This was fixed in the upstream kernel in 3ba41621156681afcdbcd624e3191cbc65eb94f4

### DIFF
--- a/kconfig/conf.c
+++ b/kconfig/conf.c
@@ -390,7 +390,7 @@ static int conf_choice(struct menu *menu)
 		}
 		if (!child)
 			continue;
-		if (line[strlen(line) - 1] == '?') {
+		if (line[0] && line[strlen(line) - 1] == '?') {
 			printf("\n%s\n", child->sym->help ?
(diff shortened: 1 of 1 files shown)


landley:
I'm in the process of replacing the kconfig plumbing (see commit 
185722d30ae7) and have long avoided patching the kconfig directory 
anyway because A) it doesn't ship, B) is grandfathered in but otherwise 
under a license I don't allow in the project.

Is this preventing your from building on some system? Security-wise I'm 
pretty sure there plenty of CFLAGS='$(rm -rf blah)' things you could 
probably do if you really tried. The build should never run this file on 
untrusted inputs.

fmayer:
I ran into this trying to build something, yes. For now I have just disabled array-bounds sanitizer on this file..

I agree that this is not a really a big security problem, but it could theoretically lead to this branch being executed by accident (if there happens to be a '?' at line[-1]), so I thought I'd fix it.

PR #547, opened 2025-04-16
dhcp: fix message type handling

dimka-rs:
get_option_msgtype() returns uint8_t, but var is defined as uint32_t. memcpy() overwrites just one byte leaving garbage in the remaining 3 bytes. This causes client to ignore valid server responses.


landley:
It should not be doing memcpy at all. And has exactly one caller...

Does commit 811d660743c9 fix it for you?

PR #581, opened 2025-11-03
sh: errexit support

alganet:
- no pipefail yet
 - no special interactive handling (but it seems to work fine)
 - added a simple fix for `set -o <flagname>` in order to test it
 - ~recursion tests were failing, added a fix~
 - added tests for a bunch of use cases


alganet:
I updated the initial idea to be more in line with the existing code, moved errexit checks to parse time, and removed the attempt at fixing the recursion stuff (realized those are probably unfinished ideas I have no business messing with).

PR #598, opened 2026-02-27
diff: Make flags with arguments non-collated

Colecf:
They should be non-collated to match gnu diff's behavior of allowing a space between the flag and its argument.

Disclaimer: this cl was written by AI. (but manually tested by me)

Fixes #597


Colecf:
Rob handled it in other commits.

PR #601, opened 2026-03-27
Always pass an access mode to open()/openat()

tavianator:
On Linux, O_RDONLY is 0, so it's used implicitly, but that's not true on
all platforms.  POSIX requires an access mode to be given explicitly.


landley:
On 3/27/26 15:05, Tavian Barnes wrote:
> On Linux, O_RDONLY is 0, so it&#39;s used implicitly, but that&#39;s not true on
> all platforms.  POSIX requires an access mode to be given explicitly.

What platforms is it not true on? It's 0 on MacOS and FreeBSD.

tavianator:
I ran into this building for WASI.  It's also true on GNU Hurd I believe

PR #604, opened 2026-06-16
fix brctl memset size

wt-z0:
`MAX_BRIDGES` is a number of elements while the third parameter of `memset()` is a byte count.

### DIFF
--- a/toys/pending/brctl.c
+++ b/toys/pending/brctl.c
@@ -47,7 +47,7 @@ static void get_ports(char *bridge, int *indices)
   unsigned long args[4] = { BRCTL_GET_PORT_LIST,
     (unsigned long) ifindices, MAX_BRIDGES, 0 };
 
-  memset(ifindices, 0, MAX_BRIDGES);
+  memset(ifindices, 0, MAX_BRIDGES * sizeof(*ifindices));
   args[1] = (unsigned long)ifindices;
   xstrncpy(ifr.ifr_name, bridge, IFNAMSIZ);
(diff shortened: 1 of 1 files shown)


enh-google:
or `int ifindices[MAX_BRIDGES] = {};` ... we're assuming a new enough -std= now, no?

(it looks like it's used in several places already, though admittedly all in code that i probably touched at some point.)

landley:
On 7/7/26 14:42, enh-google wrote:
> enh-google left a comment (landley/toybox#604)
> 
> or `int ifindices[MAX_BRIDGES] = {};` ... we're assuming a new enough -std= now, no?

C11 section 6.7.9.11 says if there aren't enough elements in the 
initializer the remaining ones are initialized as if static, meaning 
same as globals, which the ELF spec says initializes them to zero. 
(Dunno why they can't just come out and say what they mean...)

Which -std= did you consider this? Because I think gcc was handling that 
fine back in c99... yup, just dug up Knoppix 3.6 from 2004 with gcc 
3.3.4 on it (copyright 2003 ia ftaghn stallman ia) and a quick hello 
world says it understood int poing[14] = {}; just fine.

So either it's C11 or it's a gcc extension from the 32 bit single 
processor days.

Rob

enh-google:
huh, yeah, even c99 6.7.8.21 says the same.

the compiler definitely used to complain about this (and hasn't for a while) ... maybe just one of those stupid overly-pedantic-doesn't-actually-matter warnings?

